PR #23, opened 2024-09-17
fix: make port configurable via env

dsiebel:
At least on my machine, using port 5000 doesn't work, because of another service already using this port.
Since the flask port is already configurable via `FLASK_RUN_PORT`, this configuration could be exposed to the user as well.

Alternative, not so serious suggestion for the port number: `8503`, that spells "EOSB" backwards.
The "8" is just there to put it in the range of HTTP ports and because allocating anything < 1024 is not advised.


.env
```diff
@@ -1,4 +1,6 @@
 EOS_VERSION=latest
+EOS_PORT=8000
```
dsiebel, line 2:
```suggestion
EOS_PORT=8503
```

danimo:
Looks good. I'd go for 3000 (see `fix/default`branch) or 8000 by default. Not hard-to-remember random ports.

dsiebel:
> Looks good. I'd go for 3000 (see `fix/default`branch) or 8000 by default. Not hard-to-remember random ports.

As I mentioned in #25, port 3000 is the default for Grafana, Express (NodeJS), MongoDB and more. Hence my suggestion to use `8503`. 
But I am not emotionally attached to any port number so I changed it (and the default) to 3000 (f5ee200).

danimo:
As mentioned in my edit, all I care for is consistency. Let's go for your port. I retract my port 3000 idea. Sorry for the confusion.

dsiebel:
> As mentioned in my edit, all I care for is consistency. Let's go for your port. I retract my port 3000 idea. Sorry for the confusion.

No worries. Changes pushed.

danimo:
@drbacke Note, mit diesem Merge ändert sich der Default-Port!

PR #35, opened 2024-09-20
Update class_heatpump.py

NormannK:
initial clean up. comments translated, parameters pulled top the top


fortrieb (approved):
Translation, removed unused code, add static class attributes

PR #46, opened 2024-09-20
Create test_flask_server.py

NormannK:
initial test setup to check the function of flask_server.py


NormannK:
idk what hapend here

PR #48, opened 2024-09-21
Update config.example.py

NormannK:
simplifications

### DIFF
--- a/config.example.py
+++ b/config.example.py
@@ -1,13 +1,14 @@
 from datetime import datetime, timedelta
 
+# Configuration
+prediction_hours = 48
+optimization_hours = 24
+strafe = 10
 
-prediction_hours=48
-optimization_hours=24
-strafe=10
-moegliche_ladestroeme_in_prozent = [0.0 ,6.0/16.0, 7.0/16.0, 8.0/16.0, 9.0/16.0, 10.0/16.0, 11.0/16.0, 12.0/16.0, 13.0/16.0, 14.0/16.0, 15.0/16.0, 1.0 ]
+# Lade Ströme in Prozent (possible charge currents in percentage)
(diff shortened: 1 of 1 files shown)


config.example.py
```diff
@@ -1,28 +1,26 @@
...
+prediction_hours = 48
+optimization_hours = 24
+strafe = 10
 
-prediction_hours=48
-optimization_hours=24
-strafe=10
-moegliche_ladestroeme_in_prozent = [0.0 ,6.0/16.0, 7.0/16.0, 8.0/16.0, 9.0/16.0, 10.0/16.0, 11.0/16.0, 12.0/16.0, 13.0/16.0, 14.0/16.0, 15.0/16.0, 1.0 ]
+# Lade Ströme in Prozent (possible charge currents in percentage)
+moegliche_ladestroeme_in_prozent = [i / 16.0 for i in range(6, 17)] # [0.0 ,6.0/16.0, 7.0/16.0, 8.0/16.0, 9.0/16.0, 10.0/16.0, 11.0/16.0, 12.0/16.0, 13.0/16.0, 14.0/16.0, 15.0/16.0, 1.0 ]
```
fortrieb, line 9:
```python
>>> len(moegliche_ladestroeme_in_prozent)
12
>>> len([i / 16.0 for i in range(6, 17)])
11
```
The `0.0` is missing.

NormannK: currently we only use the maximum value of that list.
What is the use case for this list? I suggest we remove completely and only save the maximum loading charging current. Also i would save it as W and not A.

SREbuilt: the (Lade)Strom SI-Unit is Amperes.
having different Ladeströme (Charging currents) will make sense, once the PV predicts less production to allow charging with full. the maximum current, but only with a lower current...
this feature is widely appreciated by wallbox owners that allow PV Überschuss-Laden (PV surplus charging), i.e with only 8 amperes out of 16 possible amperes. The benefit from doing so is to avoid or minimize buying/importing electricity (Netzbezug vermeiden/minimieren).
Especially in the seasons of autumn and spring, but also in summer this helps a lot to maximize charging time per day.

NormannK: thanks for your input @SREbuilt but this might get lost here in this one month old PR :-)
if you would bring that up in the discussion that would be wonderful.
And I also agree to add it into the backlog as a feature request.

PR #49, opened 2024-09-21
Create config.example.ini

NormannK:
stump implementation of a config file to be used with  https://docs.python.org/3/library/configparser.html including some examples how to get the values
pros:
- config is not executable
- easy for user
- cross-language compatibility
- dynamic changes during runtime possible 
cons:
- complicated
- no inline comments
- Validation and Error Handling
- no logic

### DIFF
--- a/config.example.ini
+++ b/config.example.ini
@@ -1,0 +1,25 @@
+[settings]
+prediction_hours = 48
+optimization_hours = 24
+strafe = 10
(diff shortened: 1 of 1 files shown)


fortrieb:
Ok, good starting point. But config file should be parsed and used inside the project, instead of adding another file to the repo which isn't helpful at the moment.

Maybe MR can be marked as draft and we can iterate on that to reach usable configuration?

NormannK:
ah very good. I did not know i can do that in my own. 
But is this type of config ok or should we use something different?

PR #73, opened 2024-10-01
Update requirements.txt

NormannK:
only what we rly need.
with removing of class_battery_soc_predictor we don't need tensorflow anymore. which is reducing the ram req for installation extremely!


michaelosthege (approved):
IMHO this is the most important PR at this point!

Dependabot already started spamming PRs, but with just these few dependencies there'll be no need for specialized dependency management tools.

danimo:
Can you rebase the commit again? I'll approve then.

NormannK:
sry im a noob. ill make a new PR.

PR #76, opened 2024-10-02
Configure pre-commit to enforce code style automatically

michaelosthege:
Hi @drbacke !
zunächst einmal: Tolles Projekt!
Mir fehlt zwar die Technik zur Anwendung, aber ich versuche mal ein bisschen mit Refactoring zu helfen.

Damit möglichst wenige andere PRs rebased werden müssen, empfehle ich frühzeitig einen [`pre-commit`](https://pre-commit.com/) einzurichten mit dem ein Code Style forciert wird.
Hier ist ein super Tutorial: https://calmcode.io/course/pre-commit/the-problem

Derzeit sind in der Python-Welt dafür `black` oder `ruff` populär.
Ruff ist schneller und übernimmt dabei noch die Funktion von `pylint` um andere Bugs zu entdecken.

-------------------

### Changes in this PR

* A `.pre-commit-config.yaml` is added alongside a GitHub action to run code style enforcements automatically.
* Developers will have to do one of the following
  * write perfect code style
  * run `pre-commit install` to run formatting automatically before every commit
  * run e.g. `pre-commit --all` manually before a commit

I made three commits to make code review easier:
1. Added the GitHub action and configuration files
2. Applied automatic formatting by running `pre-commit run --all` and committing the changes
3. Manually edited some parts to get rid of remaining errors. Mostly:
  * Avoiding `from stuff import *` imports because the linters can't deal with them
  * Removing unused imports
  * Renamed `config.example.py` → `config.py` to make imports work (in theory). Switching to `configparser` (#49) is a much better way to do it.

------------------

### Folgearbeiten

Ich würde außerdem empfehlen eine `CONTRIBUTING.md` anzulegen und dort die Hinweise aus der README und die o.g. Hinweisen bezüglich `pre-commit` zusammenzufassen.


config.py
fortrieb, line 1: We should make it more clear users needs be editing this file to get startet and i think the file extension `config.example.py` could be helpful to navigate a new user where to get started.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+- repo: https://github.com/PyCQA/isort
+  rev: 5.13.2
+  hooks:
+    - id: isort
+      name: isort
+- repo: https://github.com/astral-sh/ruff-pre-commit
+  rev: v0.6.8
+  hooks:
+    # Run the linter and fix simple issues automatically
+    - id: ruff
```
fortrieb, line 21: Did a developer needs `ruff` installed? If so, this should be also added as a dependency.

fortrieb (changes requested):

config.py
michaelosthege: I updated the corresponding entries in both READMEs.

However, I would strongly recommend to move config options into an INI or JSON file, with the option to set some of them (e.g. passwords) via environment variables or command line arguments.

.pre-commit-config.yaml
michaelosthege: No, only `pre-commit`. I can open a follow-up PR to summarize this in a `CONTRIBUTING.md`.

Usually the development dependencies go into something like a `requirements-dev.txt` because they are not needed at runtime. (I can make a PR for that too.)

fortrieb: Yes, dev requirements sound good.

noootch: I would suggest switching to poetry this would enable using one single pyproject.toml for dependencies and other important configuration like versioning or e.g. ruff settings

config.py
noootch: I heard you: https://github.com/Akkudoktor-EOS/EOS/pull/79

.pre-commit-config.yaml
michaelosthege: pyproject.toml as the single point for configuration and package metadata works well without poetry!
After reducing the number of dependencies I don't see why one would need to introduce an advanced dependency management tool that is rarely used by beginners.

noootch: Or without poetry.

NormannK:
i love this PR. would be awesome if it would be run on all the code right away

drbacke:
The changes are enormous, all files are affected. 
I currently find this very difficult, as I seem to be the only one who can test the code at all. I would therefore like to finish the unit tests first, otherwise I will only one testing PRs. It can't go on like this

noootch:
it's just formatting changes, isn't it?

I think the earlier some code formatting rules are implemented the easier maintaining and developing will be. When the code base grows and rules are implemented at a later stage the tech dept will grow and developing will become harder.

drbacke:
That may be, but I still can't check all functional pr alone, it's a general problem. 
I can't get any further with the physical/technical development. Nobody uses the code except me and that's a very bad basis. Most of the programming here is done on speculation. 
I think the current development is extremely questionable

michaelosthege:
> That may be, but I still can't check all functional pr alone, it's a general problem. I can't get any further with the physical/technical development. Nobody uses the code except me and that's a very bad basis. Most of the programming here is done on speculation. I think the current development is extremely questionable

I agree that unit tests should be added before making any functional changes.

Still, it will be much easier to review and add tests with some basic automated tests. The ones from this PR, for example, will already catch things like typos in variable names, syntax errors and so on.

I'm happy to walk you through these changes. I've been through this in countless other projects.

drbacke:
> > That may be, but I still can't check all functional pr alone, it's a general problem. I can't get any further with the physical/technical development. Nobody uses the code except me and that's a very bad basis. Most of the programming here is done on speculation. I think the current development is extremely questionable
> 
> I agree that unit tests should be added before making any functional changes.
> 
> Still, it will be much easier to review and add tests with some basic automated tests. The ones from this PR, for example, will already catch things like typos in variable names, syntax errors and so on.
> 
> I'm happy to walk you through these changes. I've been through this in countless other projects.

Yes, that's a good point. I like to check the pr and can also do a quick test. 
Let's continue the discussion regarding the tests / active users of the tool in the discussions

PR #78, opened 2024-10-02
Pytest for Heatpump

fortrieb:
This PR will implement a unit tests for class `Heatpump`.


drbacke:
Good serve, I also wrote a first unit test for the battery class and then integrate my tests into your environment.
However, the tests from the physical side are still rather nonsensical, but I think this is about the software side for now.

PR #79, opened 2024-10-02
Improve config

noootch:
My first contribution : )

I updated the configuration handling. The configuration is now done via json file.
I tried to find all spots where the configuration is used.

This PR includes:

- new configuration handling
- testing of the configuration handling
- some vscode improvements

### Sharing vscode configurations and settings
Sharing vscode settings with the suggested extension allows users to create some local settings. Project specific configuration is put in xyz.shared.json, local settings can be added in xyz.local.json. The plugin merges both files to xyz.json which is then used by vscode. This is especially helpful with launch.json.

### Pydantic
With pydantic the config is type safe and the data is validated. Also it improves the workflow a lot when you can directly access the config attributes.


drbacke:
Please check the conflicts

PR #81, opened 2024-10-03
Add some tools

noootch:
- added mypy and ruff as preparation for testing and forcing some code style.
- ran _**ruff format**_
- added workspace config with some settings and extension recommendations

I can give a short introduction on the improved workflow.


modules/class_pv_forecast.py
```diff
@@ -54,51 +67,54 @@ def __init__(self, filepath=None, url=None, cache_dir='cache', prediction_hours=
...
         input_date_hour = date_time.replace(minute=0, second=0, microsecond=0)
 
         for forecast in self.forecast_data:
-            forecast_date_hour = parser.parse(forecast.date_time).replace(minute=0, second=0, microsecond=0)
+            forecast_date_hour = parser.parse(forecast.date_time).replace(
+                minute=0, second=0, microsecond=0
+            )
             if forecast_date_hour == input_date_hour:
                 forecast.ac_power_measurement = ac_power_measurement
                 found = True
```
NormannK, line 84: Local variable `found` is assigned to but never used

config.example.py
```diff
@@ -1,28 +1,37 @@
...
-        else:
-                date = (startdate + timedelta(hours = prediction_hours)).strftime("%Y-%m-%d")
-                date_now = startdate.strftime("%Y-%m-%d")
-        return date_now,date
+
+def get_start_enddate(prediction_hours=48, startdate=None):
+    ############
+    # Parameter
+    ############
+    if startdate == None:
```
NormannK, line 30: startdate is None

NormannK (approved):

modules/class_pv_forecast.py
NormannK: i guess ruff is not looking for that. i just noticed while looking into the code. to keep the pr clean lets keep it for now

pyproject.toml
```diff
@@ -1,7 +1,29 @@
...
+[tool.pyright]
+# used in Pylance extension for language server
+# type check is done by mypy, disable to avoid unwanted errors
+typeCheckingMode = "off"
+
 [tool.isort]
 profile = "black"
 
+[tool.ruff]
+line-length = 100
```
noootch, line 23: my suggestion is to extend the line length as modern dev setups usually don't have a low display limit.
Also the code is less dense with less line breaks (even though the code in each line should be short anyways.

fortrieb: +1

requirements-dev.txt
fortrieb, line 1: Should `mypy` be added to CI pipeline? I don't using `mypy` and not I'm not aware of there capabilities.

fortrieb (changes requested):

requirements-dev.txt
michaelosthege: We use mypy in several projects, but this one is far from the point where adding it makes sense.
Some important dependencies such as sklearn and pandas are not great when it comes to type hints, and the code here doesn't have any type hints either.

IMO what's needed first is a refactoring that structures the code into submodules (devices, optimization, io, ...) and a switch to dataclasses with type annotations and docstrings instead of dicts.

noootch: mypy helps during development as in combination with the extension it highlights issues in the code

flask_server.py
```diff
@@ -311,13 +296,14 @@ def root():
     return redirect("/site-map", code=302)
 
 
-if __name__ == "__main__":
+def run_flask(default_host: str = "0.0.0.0", default_port: int = 8503) -> None:
```
noootch, line 299: I moved the default port to the method call for now. @fortrieb 

If don't know if the flask setup should be put in a config json here. This PR focuses on introducing ruff rules. And the rule I fixed here was "no magic number".

drbacke:
Pre commit does that already. Aren't These Changes outdated?

noootch:
This PR is a vscode setup. It enables syntax highlighting so that you can see errors
while programming and not only when you try to commit. I can add a screenshot later.

The formatting changes come from longer lines. Default 79 -> 100. I wrote my reason for this in the specific line in the toml file.

noootch:
Example ruff:
grafik

Example mypy:
grafik

requirements-dev.txt
```diff
@@ -1,2 +1,6 @@
 pytest==8.3.3
+mypy==1.11.2
+ruff==0.6.8
+isort==5.13.2
+
```
michaelosthege, line 5: They are not needed as explicit dependencies. The ruff and isort versions are already defined by `.pre-commit-config.yaml`.

To run them follow the instructions in [CONTRIBUTING.md](https://github.com/Akkudoktor-EOS/EOS/blob/main/CONTRIBUTING.md#code-style)

modules/heatpump.py
```diff
@@ -32,7 +32,7 @@ def __check_outside_temperature_range__(self, temp_celsius: float) -> bool:
         Returns:
             bool: True if in range
         """
-        return temp_celsius > -100 and temp_celsius < 100
+        return temp_celsius > -100 and temp_celsius < 100  # noqa: PLR2004
```
michaelosthege, line 35: There will always going to be a lot of magic values in this project. At this point it would complicate things to ask developers to put magic comments at them.

It's more important to get configuration and architecture right.
Automated checks saying "don't do this" without good architecture for how to do it better may create a mess.

pyproject.toml
```diff
@@ -1,7 +1,29 @@
...
+
 [tool.ruff.lint]
 ignore = [
     "F841",  # don't complain about unused variables
 ]
+select = ["RUF", "FURB", "PL"] # suggestion: add PTH
```
michaelosthege, line 29: These are quite strict, and some are questionable too (e.g. FURB140).

At this point I would focus attention on the bigger issues (code structure) instead of small details such as variable naming.
```suggestion
select = ["RUF"]
```

EOS.code-workspace
```diff
@@ -0,0 +1,29 @@
...
+    "recommendations": [
+      "ms-python.mypy-type-checker",
+      "ms-python.python",
+      "ms-python.debugpy",
+      "charliermarsh.ruff",
+      "esbenp.prettier-vscode",
+      "ms-python.isort"
+    ]
+  }
+}
```
michaelosthege, line 29: Why should the project have this file at the root level?

michaelosthege (commented):
I like some things in here. Can you extract _only the max line length change_ to a separate PR?
This will make it easier to review and rebase other PRs.

modules/heatpump.py
noootch: removed the rules

requirements-dev.txt
noootch: done

EOS.code-workspace
noootch: removed workspace config and moved changes to settings.shared and extensions

pyproject.toml
noootch: removed

PR #85, opened 2024-10-03
feat: run pytest for PRs

danimo:


.github/workflows/pytest.yml
```diff
@@ -0,0 +1,28 @@
...
+          python-version: '3.12'
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+
+      - name: Run Pytest
+        run: |
+          python -m pytest -v tests/
```
michaelosthege, line 28:
```suggestion
          pytest -vs --cov modules --cov-report term-missing tests/
```
may need `pytest-cov` as additional dependency

danimo: running pytest yields "modules not found". hence the "python -m" approach. Still I can add the remaining parameters.

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,29 @@
...
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
```
michaelosthege, line 18: the action versions are outdated. See https://github.com/Akkudoktor-EOS/EOS/blob/156eddc6ef878faae84ceba7b0997a269e64d197/.github/workflows/pre-commit.yml#L13

PR #86, opened 2024-10-03
WIP Fix Docker setup

Lasall:
pandas was missing from the deps, also I had to change the Wechselrichter import. The mariadb connection config should probably support env variables so no manual change is required to start it with compose (I can add it to this PR). However the table update in the /soc endpoint fails due to the table being never created. How is it supposed to look like?

I would suggest to use the Docker build + tests/startup as pipeline configuration. The image could be exported to dockerhub. I usually work with GitLab, but if you agree with those suggestions I could set them up for Github.


drbacke:
Please check the conflicts

fortrieb:
Would be great to integrate docker build into Github CI pipeline. An export to dockerhub isn't needed currently, we have to improve the project before publishing it.

drbacke:
> Would be great to integrate docker build into Github CI pipeline. An export to dockerhub isn't needed currently, we have to improve the project before publishing it.

Good Idea👍 Would be much easier for using it

drbacke:
> pandas was missing from the deps, also I had to change the Wechselrichter import. The mariadb connection config should probably support env variables so no manual change is required to start it with compose (I can add it to this PR). However the table update in the /soc endpoint fails due to the table being never created. How is it supposed to look like?
> 
> I would suggest to use the Docker build + tests/startup as pipeline configuration. The image could be exported to dockerhub. I usually work with GitLab, but if you agree with those suggestions I could set them up for Github.

Could you resolve the remaining conflicts?
Next step Would be a GitHub Action to Export the docker File.

Lasall:
Sorry, was away over the weekend, will be back later today. I will add an env for the DB connection so the compose works ootb.
Will discussions about the project happen here or on the forums? For example if a database is used I would recommend an ORM approach (flask+SQL alchemy+alembic or directly switch to django). Is user management a future requirement (-> django) etc.

michaelosthege:
EOS currently uses exactly one table and makes three different queries to it.

Is a database even needed?

I created a discussion: https://github.com/Akkudoktor-EOS/EOS/discussions/108

drbacke:
No, the EOS consists of the optimization algo. and numerous “ secondary tools”.
The ancillary tools such as PV forecast or load forecast are currently (partially) still linked to my DB system. This must be changed in the future, and a DB connection should be avoided entirely.

PR #88, opened 2024-10-04
Refactored class_optimize.py

NormannK:
- Optimized Imports: Removed unused imports and organized them.
- Refactored Code: Introduced split_individual function for clarity.
- Improved Efficiency: Enhanced penalty calculation and streamlined loops.
- Updated Evaluation Logic: Better handling of penalties in evaluate.
- Type Hints added.

Notes:
- Fixed seed option added for automated tests
- Verbose option added, default False
- isfloat is only used in flask_server.py
- start_hour is not used in this class
- identical results with the old version using test.py and fixed seeds

### DIFF
--- a/modules/class_optimize.py
+++ b/modules/class_optimize.py
@@ -1,25 +1,23 @@
 import os
+import random
 import sys
+from typing import Any, Dict, List, Optional, Tuple
 
-import matplotlib
 import numpy as np
+from deap import algorithms, base, creator, tools
 
 from modules.class_akku import PVAkku
-from modules.class_ems import EnergieManagementSystem, Wechselrichter
+from modules.class_ems import EnergieManagementSystem
 from modules.class_haushaltsgeraet import Haushaltsgeraet
+from modules.class_inverter import Wechselrichter
 from modules.visualize import visualisiere_ergebnisse
-
-matplotlib.use("Agg")  # Setzt das Backend auf Agg
-import random
-from datetime import datetime, timedelta
-
-from deap import algorithms, base, creator, tools
 
 sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 from config import moegliche_ladestroeme_in_prozent
 
 
-def isfloat(num):
+def isfloat(num: Any) -> bool:
+    """Check if a given input can be converted to float."""
     try:
         float(num)
         return True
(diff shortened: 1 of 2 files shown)


NormannK:
statistically proof that the new class is as good as (the same) as the old one.
image
see #93 for results from the original class

modules/class_optimize.py
```diff
@@ -373,121 +282,71 @@ def optimierung_ems(
...
             eautocharge_hours_float,
-            temperature_forecast,
+            parameter["temperature_forecast"],
             start_hour,
             self.prediction_hours,
             einspeiseverguetung_euro_pro_wh,
             extra_data=extra_data,
         )
+
         os.system("cp visualisierungsergebnisse.pdf ~/")
```
fortrieb, line 341: Result PDF path file should be configurable, maybe a method parameter or config parameter.

modules/class_optimize.py
```diff
@@ -1,307 +1,233 @@
...
+            if attr in creator.__dict__:
+                del creator.__dict__[attr]
 
+        # Create new FitnessMin and Individual classes
         creator.create("FitnessMin", base.Fitness, weights=(-1.0,))
         creator.create("Individual", list, fitness=creator.FitnessMin)
 
-        # PARAMETER
+        # Initialize toolbox with attributes and operations
         self.toolbox = base.Toolbox()
```
fortrieb, line 88: Initialization should be move to `__init__`. Toolbox is used in other places as well.

modules/class_optimize.py
```diff
@@ -1,307 +1,233 @@
...
-        self.prediction_hours = prediction_hours  #
+    def __init__(
+        self,
+        prediction_hours: int = 24,
+        strafe: float = 10,
+        optimization_hours: int = 24,
+        verbose: bool = False,
+        fixed_seed: Optional[int] = None,
+    ):
+        """Initialize the optimization problem with the required parameters."""
```
fortrieb, line 37: Docstring can be more helpful to describe each parameter. I'm using [Google Style docstrings](https://google.github.io/styleguide/pyguide.html#38-comments-and-docstrings), maybe we can discuss on using in this project.

drbacke: Lets discuss this in the discussion section.

PR #89, opened 2024-10-04
Heat pump: Tests, Translation, Docstrings, Relocation

fortrieb:
Heat pump class is refactored for this PR. Following things where done:

- Translation to english
- Adding type hints to methods
- Check inputs for methods
- Adding Python doc strings to methods
- Add more units tests to cover all methods
- Remove `class` prefix of python file


michaelosthege:
Run these to fix code style:
* `pip install -r requirements-dev.txt`
* `pre-commit run --all`

modules/heatpump.py
```diff
@@ -0,0 +1,133 @@
+import logging
+from typing import List
+
+
+class HeatPump:
+    MAX_HEATOUTPUT = 5000  # Maximum heating power in watts
+    BASE_HEATPOWER = 235.0  # Base heating power value
+    TEMPERATURE_COEFFICIENT = -11.645  # Coefficient for temperature
+    COP_BASE = 3.0  # Base COP value
+    COP_COEFFICIENT = 0.1  # COP increase per degree
```
michaelosthege, line 10: This is the docstring syntax for class attributes:

```suggestion
    MAX_HEATOUTPUT = 5000
    """Maximum heating power in watts"""

    BASE_HEATPOWER = 235.0
    """Base heating power value"""

    TEMPERATURE_COEFFICIENT = -11.645
    """Coefficient for temperature"""

    COP_BASE = 3.0
    """Base COP value"""

    COP_COEFFICIENT = 0.1
    """COP increase per degree"""
```

requirements-dev.txt
```diff
@@ -1,2 +1,2 @@
 pytest==8.3.3
-pre-commit
+pre-commit==3.8.0
```
michaelosthege, line 2: For some dependencies the exact versions don't matter and pinning them can even increase maintenance workload:
* Dependabot making PRs about every small version increment that doesn't actually matter
* Devs won't actually `pip install -r requirements-dev.txt` every time they continue working on the project

For the `requirements.txt` I would even consider it bad practise to pin all versions exactly, because it makes it very hard to install the library into an environment also containing other packages.

Fixing it for the Docker-build, however, would be good practice (as long as it's kept updated by dependabot).

tests/conftest.py
```diff
@@ -0,0 +1,9 @@
+import pytest
+
+from modules.heatpump import HeatPump
+
+
+@pytest.fixture(scope="function")
+def hp_5kw_24h() -> HeatPump:
+    """Heat pump with 5 kw heating output and 24 h prediction"""
+    return HeatPump(5000, 24)
```
michaelosthege, line 9: why not put it in `test_heatpump` to keep the related things together?

tests/test_heatpump.py
```diff
@@ -1,17 +1,44 @@
...
-@pytest.fixture(scope="function")
-def heatpump() -> Heatpump:
-    """Heatpump with 5 kw heating power and 24 h prediction"""
-    return Heatpump(5000, 24)
+class TestHeatPump:
+    def test_cop(self, hp_5kw_24h: HeatPump):
+        """Testing calculate COP for variouse outside temperatures"""
+        assert hp_5kw_24h.calculate_cop(-10) == 2.0, "COP for -10 degree isn't correct"
+        assert hp_5kw_24h.calculate_cop(0) == 3.0, "COP for 0 degree isn't correct"
+        assert hp_5kw_24h.calculate_cop(10) == 4.0, "COP for 10 degree isn't correct"
```
michaelosthege, line 11: the messages are trivial, given that the assert statement is already quite self-explanatory
```suggestion
        assert hp_5kw_24h.calculate_cop(-10) == 2.0
        assert hp_5kw_24h.calculate_cop(0) == 3.0
        assert hp_5kw_24h.calculate_cop(10) == 4.0
```

tests/test_heatpump.py
```diff
@@ -1,17 +1,44 @@
...
+        # Check edge case for outside temperature
+        out_temp_min = -100.1
+        out_temp_max = 100.1
+        with pytest.raises(ValueError) as err:
+            hp_5kw_24h.calculate_cop(out_temp_min)
+        assert str(out_temp_min) in str(err.value)
+        with pytest.raises(ValueError) as err:
+            hp_5kw_24h.calculate_cop(out_temp_max)
+        assert str(out_temp_max) in str(err.value)
```
michaelosthege, line 20:
```suggestion
        with pytest.raises(ValueError, match=f"'{out_temp_min}' not in range"):
            hp_5kw_24h.calculate_cop(out_temp_min)
        with pytest.raises(ValueError, match=f"'{out_temp_max}' not in range"):
            hp_5kw_24h.calculate_cop(out_temp_max)
```

tests/test_heatpump.py
```diff
@@ -1,17 +1,44 @@
...
+        """Testing calculate of heating output"""
+        assert (
+            hp_5kw_24h.calculate_heating_output(-10.0) == 5000
+        ), "Wrong output at -10.0 Celsius"
+        assert (
+            hp_5kw_24h.calculate_heating_output(0.0) == 5000
+        ), "Wrong output at 0.0 Celsius"
+        assert hp_5kw_24h.calculate_heating_output(10.0) == pytest.approx(
+            4939.583
+        ), "Wrong output at 10.0 Celsius"
```
michaelosthege, line 32:
```suggestion
        assert hp_5kw_24h.calculate_heating_output(-10.0) == 5000
        assert hp_5kw_24h.calculate_heating_output(0.0) == 5000
        assert hp_5kw_24h.calculate_heating_output(10.0) == pytest.approx(4939.583)
```

tests/test_heatpump.py
```diff
@@ -1,17 +1,44 @@
...
+        """Testing calculation of heating power"""
+        assert (
+            hp_5kw_24h.calculate_heat_power(-10.0) == 2104
+        ), "Wrong power at -10.0 Celsius"
+        assert (
+            hp_5kw_24h.calculate_heat_power(0.0) == 1164
+        ), "Wrong power at 0.0 Celsius"
+        assert (
+            hp_5kw_24h.calculate_heat_power(10.0) == 548
+        ), "Wrong power at 10.0 Celsius"
```
michaelosthege, line 44:
```suggestion
        assert hp_5kw_24h.calculate_heat_power(-10.0) == 2104
        assert hp_5kw_24h.calculate_heat_power(0.0) == 1164
        assert hp_5kw_24h.calculate_heat_power(10.0) == 548
```

modules/heatpump.py
```diff
@@ -0,0 +1,133 @@
+import logging
+from typing import List
```
michaelosthege, line 2:
```suggestion
from typing import List, Sequence
```

modules/heatpump.py
```diff
@@ -0,0 +1,133 @@
...
+                + 1.62 * outside_temperature_celsius**2.0
+            )
+        else:
+            err_msg = f"Outside temperature '{outside_temperature_celsius}' not in range (min: -100 Celsius, max: 100 Celsius) "
+            self.log.error(err_msg)
+            raise ValueError(err_msg)
+
+    def simulate_24h(self, temperatures: List[float]) -> List[float]:
+        """Simulate power data for 24 hours based on provided temperatures."""
+        power_data = List[float]
```
michaelosthege, line 103: Rule of thumb: Use invariant types (`Sequence`, `Collection`, `Iterable`, `Mapping`, ...) for parameters and covariant types (`List`, `Tuple`, `Dict`, ...) for return values.
This keeps the code flexible while improving the utility of type information.

In this example: `.simulate_24h(temperatures=(1,2,3,...,24))` works with the implementation, but wouldn't be allowed if `List[float]` was required for the input.

```suggestion
    def simulate_24h(self, temperatures: Sequence[float]) -> List[float]:
        """Simulate power data for 24 hours based on provided temperatures."""
        power_data: List[float] = []
```

michaelosthege (commented):
Good refactoring!

Mid-term I see the issue of defining heat pumps with different properties, or even multiple heat pumps.
But I think this is a topic that should be covered by a discussion before making implementation changes.

requirements-dev.txt
fortrieb: Thanks for the explanation.

tests/conftest.py
fortrieb: Thought we can place all fixture in one place. Maybe adding a subfolder `fixtures` inside `tests` which can be used for that.

michaelosthege (approved):

PR #98, opened 2024-10-04
test for class_optimize in PR #88

NormannK:
Test meant for class_optimize.
Will only work with PR #88 since the old class has no way to use a fixed random seed.

To use it on the old/current class add `random.seed(42)` to the __init__ of  `optimization_problem` and  add `fixed_seed=None` as a possible argument.

pre-commit style of the lists and dicts does look odd.
It takes quite a while to finish also there are two warnings from the visualization class. 
Suggestions are appreciated.


danimo:
can you fold this into #88 then?

NormannK:
Sure I can, but im not so sure its the right way to do it. If you say its the proper way to do it id be happy to do so.

danimo:
Well if the test doesn't work prior to #88, but tests #88, then it should go with #88. Maybe also rebase #88 on top of main if you know how to do so.

NormannK:
> Well if the test doesn't work prior to #88, but tests #88, then it should go with #88. Maybe also rebase #88 on top of main if you know how to do so.

merged thanks! ill send this in to review

drbacke:
Please merge this into the newest code, the "wechselrichter" bug is already resolved.

michaelosthege:
that'll be
```
git fetch origin
git rebase origin/main
```

NormannK:
thank you very much @michaelosthege I just redid my whole setup and learned a lot. i should go more smooth in the future (fingers crossed)

PR #103, opened 2024-10-05
cleanup class_ems.py

NormannK:
- type hints added
- translated to English
- reduced not needed function calls 

improves optimization time by 50%


NormannK:
["Last_Wh_pro_Stunde"][0], ["Netzeinspeisung_Wh_pro_Stunde"][0], ["Netzbezug_Wh_pro_Stunde"][0] were `None ` bc the array was initialized with nan values which later got converted into None. Those but that should be numbers afaik.
please confirm.

NormannK:
old `Last_Wh_pro_Stunde` result:
`[None, 23759.13, 996.88, 531.38, 517.95, 483.15, 472.28, 1011.68, 995.0, 1053.07, 1063.91, 1320.56, 1132.03, 1163.67, 1176.82, 1216.22, 1103.78, 1129.12, 1178.71, 1050.98, 988.56, 912.38, 704.61, 516.37, 868.05, 694.34, 608.79, 556.31, 488.89, 506.91, 804.89, 1141.98, 1056.97, 992.46, 1155.99, 827.01, 1257.98, 1232.67, 871.26, 860.88, 1158.03, 1222.72, 1221.04, 949.99, 987.01, 733.99, 592.97] `
new `Last_Wh_pro_Stunde` result:
`[0.0, 23759.13, 996.88, 531.38, 517.95, 483.15, 472.28, 1011.68, 995.0, 1053.07, 1063.91, 1320.56, 1132.03, 1163.67, 1176.82, 1216.22, 1103.78, 1129.12, 1178.71, 1050.98, 988.56, 912.38, 704.61, 516.37, 868.05, 694.34, 608.79, 556.31, 488.89, 506.91, 804.89, 1141.98, 1056.97, 992.46, 1155.99, 827.01, 1257.98, 1232.67, 871.26, 860.88, 1158.03, 1222.72, 1221.04, 949.99, 987.01, 733.99, 592.97] 
`

michaelosthege:
I haven't familiarized myself with the simulation and optimization yet, but my gut feeling is that `NaN` should be prefered over `None`.
For example, a `np.array([1,None,3])` is of `dtype=object` whereas `np.array([1, np.nan, 2])` is `dtype=float`.

NormannK:
yeah i was also not that sure anymore. I read this [article ](https://www.geeksforgeeks.org/difference-between-nan-and-none-in-python/) to remember things. `NaN` is a float by definition and None is a "None Type" so dtype must change.
But for the optimization in the end that would not matter since the last step in the class was/is to translate all `NaN` into `None` anyways.

I would understand if, if the first value is `NaN` or `None` if we dont want to have effect that value the current running operation. So basically saying "hey don't touch anything for the current hour we only want to have values for the next hour" 
Be we only have no value there bc the line 103 in the old ` for stunde in range(start_stunde + 1, ende):` starts one hour later and there is no documentation why it does that. (or i cant find it)

drbacke:
Home Appliances not working anymore. -> Issues (should be checked)

drbacke (changes requested):
Home Appliances not working anymore

NormannK:
the test_class_ems.py sets `haushaltsgeraet = None`  so it assumes there are none. 

also i did check the result vs single_test_optimization.py. Its the same.
The code does not touch any logic, could you provide the input data you used?

drbacke:
I fixed it, the call (set_startzeitpunkt returns no object, that was the problem). 
I also added the home appl. to the unit tests, so that we will see that problem in the future.

Issue #107, opened 2024-10-06
Refactor optimization-tests to load input/expectation data from JSON files


michaelosthege:
### Description

Currently located in `tests/test_class_optimize.py` there are about 1000 lines of test data (input and expectation) hard-coded into the Python script.

This should be moved to JSON files, for example:
* `tests/testdata/optimize_input_1.json`
* `tests/testdata/optimize_result_1.json`

Then, the test function can be parametrized using pytest:

```python
DIR_TESTDATA = Path(__file__).parent / "testdata"


@pytest.mark.parametrize("fn_in,fn_out", [("optimize_input_1.json", "optimize_output_1.json")])
def test_optimize(fn_in, fn_out):
    # TODO: load input/expectation data
    pass
```

NormannK:
I thought about making a hash out of that result and compare it. Would that be a proper alternative? 
The only downside I see would that you cant compare what actually is different.

michaelosthege:
No that's not something you'd do with hashes, for the reason you already mentioned.

`np.testing` has a lot of useful assertion functions to compare arrays. They also report the differences.

NormannK:
assign that to me. ill to it later today.

NormannK:
@michaelosthege can you look at PR #116 please? I can't figure out why the json flies are not being found.

PR #110, opened 2024-10-06
fixed_seed for single_test_optimization

NormannK:
sets the seed for random to a fixed value to be able to compare results. Value chosen for obvious reason. Same value is used in the test case.


NormannK:
the "fixed optimize test" is unrelated to the fixed seed change. I'm working on a better test. #107

PR #123, opened 2024-10-07
remove unused function

NormannK:
replace_nan_with_none(data) is not needed anymore since we don't have any NaN values. Another 13% speed increase.


drbacke (changes requested):
13% is impossible, this function is called once. 
Also the use of zero values is not allowed, because zero is a valid value. But some results cannot be determined for the first entry. So there must be a placeholder, like None oder Null.

NormannK:
change es requested in #132 
13% is absolutely possible since we call this function ~160 times per optimization generation!
Please check if other values need to be set since we don't have a test for them or a documentation

drbacke (approved):
This should work. Thanks for changing :-)

Issue #126, opened 2024-10-07
[BUG]: isfloat accepts no None


drbacke:
### Describe the issue:

flask_server.py  isfloat function does not accept None, should return False

### Reproduceable code example:

```python
isfloat(None)
```

### Version information:

-

NormannK:
this is resolved

PR #127, opened 2024-10-07
improve and move isfloat(num)

NormannK:
fixing #126 by adding case "None", stings with whitespace and special types like +-inf.
Also moving the function to the server since its only used there.

### DIFF
--- a/src/akkudoktoreosserver/flask_server.py
+++ b/src/akkudoktoreosserver/flask_server.py
@@ -2,6 +2,7 @@
 
 import os
 from datetime import datetime
+from typing import Any
 
 import matplotlib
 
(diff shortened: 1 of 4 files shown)


src/akkudoktoreosserver/flask_server.py
```diff
@@ -26,6 +27,25 @@
 )
 
 
+def isfloat(num: Any) -> bool:
```
noootch, line 30: you could add a TypeGuard:
`
from typing import TypeGuard
def isfloat(num: Any) -> TypeGuard[float]
`

NormannK: oh that's a very nice function I did not know about. 
According to PEP 647 thats in Version 3.10 introduced. Currently we need 3.8+ according to the readme. Would 3.10+ be a problem? Ill add it

noootch: I did not check the required python version, my bad.

NormannK: do you see any reason not to change the python version needed? 3.10 is already 4 years old.

noootch: personally I would even go to 11 or even 13. But as the program is also distributed as a package that other people can implement, it is better to support more versions.

NormannK:
@drbacke 

I guess I'm the only one besides you running that code actively. In my backend I save everything into a table and go from there. Also I simply ignore the first value. 

But i see what you are referring too. So we don't have proper type sanitation before translating it into json.
Since the Flask server can't call the EMS the only way to hand the result over must go through the optimization. 
So the right way would be to translate everything which comes out of the optimization class into lists where the None value can be translated into the equivalent "Null" json value. 
Im looking into the code of the flask_server for "gesamtlast_simple" we do a `return jsonify(last.tolist()) `
But for "optimize" there is a simple return jsonify(result).
So the right way to do it take care of the translations at the last step where we do the conversion and not in between. 
Also if that fails we miss a test case....

noootch:
Pydantic is also a good option to parse data from and to json. The bonus point: you have a nice object that carries your data so you can access it via attributes which increases type safety.

I used it in my suggestion for the config.

@drbacke regarding the config, can you give me a feedback why you closed the PR?

michaelosthege:
@noootch you are referring to https://github.com/Akkudoktor-EOS/EOS/pull/79. I really liked your proposal to use `pydantic` to data validation.

I'm 75 % sure it got closed automatically when @drbacke accidentally force-pushed the `main` branch to a new history (same commits, new hashes from day 0) 👇

image

noootch:
meeeeh, ok. I'll reopen with the new branch.

PR #129, opened 2024-10-07
Streamline Dockerfile, remove unused deps and code

Lasall:
* Dockerfile: Use non-root user, buildx cache, setup for readonly container, remove unused apt deps. For now don't install pip package and keep development flask server as this will be replaced in the future (fastapi). Then a proper webserver (e.g. nginx) should be used and the pip package can be created and deployed just to the run-stage (with the webserver).
 * docker-compose: Set to readonly (anonymous volumes declared in Dockerfile should maintain all writable data). Mount config.py for easier development. Should be replaced by environment support for all config file variables.
 * Remove now unused class_soc_calc.
 * Remove unused runtime dependencies: mariadb, joblib, pytest, pytest-cov.
 * Move pytest-cov to dev dependencies.
 * Update docs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,7 +24,7 @@ To make collaboration easier, we require pull requests to pass code style and un
 Our code style checks use [`pre-commit`](https://pre-commit.com).
 
 ```bash
-pip install -r requirements.txt
+pip install -r requirements-dev.txt
 ```
 
 To run formatting automatically before every commit:
@@ -36,7 +36,7 @@ pre-commit install
 Or run them manually:
 
 ```bash
-pre-commit --all
+pre-commit run --all-files
 ```
 
 ### Tests
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -20,9 +20,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          sudo apt install -y libmariadb3 libmariadb-dev
           python -m pip install --upgrade pip
-          pip install -r requirements.txt
           pip install -r requirements-dev.txt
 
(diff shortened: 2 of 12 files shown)


CONTRIBUTING.md
```diff
@@ -36,7 +36,7 @@ pre-commit install
 Or run them manually:
 
 ```bash
-pre-commit --all
+pre-commit run
```
michaelosthege, line 39: this runs pre-commit only on staged files and I wouldn't assume that beginner contributors know the difference..

src/akkudoktoreos/class_soc_calc.py
michaelosthege, line 1: Thanks for pointing out that it's unused, but better not remove it in this PR.

We didn't have a chance to sync up on architecture roadmap and the `BatteryDataProcessor` might become relevant again.

michaelosthege (commented):
Good stuff about the Dockerfile and updating the instructions!

CONTRIBUTING.md
Lasall: actually i don´t know how it worked, I just tried to run it and it didn´t do anything:
```
usage: pre-commit [-h] [-V] {autoupdate,clean,gc,init-templatedir,install,install-hooks,migrate-config,run,sample-config,try-repo,uninstall,validate-config,validate-manifest,help,hook-impl} ...
pre-commit: error: unrecognized arguments: --all
```

michaelosthege: Oh right, it should say
```suggestion
pre-commit run --all-files
```

(but `pre-commit run --all` works too)

michaelosthege (approved):

PR #139, opened 2024-10-08
Add settings and extension recommendations

noootch:
A suggestion for some vscode configuration to improve the workflow.

# Pytest
grafik

# TODO
- [x] Add debug configuration
- [x] Test test debugging
- [x] Add arg to makefile

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,14 @@
+{
+  "recommendations": [
+    //python
+    "ms-python.python",
+    "ms-python.debugpy",
+    "charliermarsh.ruff",
+    "ms-python.mypy-type-checker",
+
+    // misc
+    "swellaby.workspace-config-plus", // allows user and shared settings
+    "christian-kohler.path-intellisense",
+    "esbenp.prettier-vscode"
+  ]
+}
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,11 @@ include = ["akkudoktoreos", "akkudoktoreosserver", ]
 [tool.setuptools.package-data]
 akkudoktoreosserver = ["data/*.npz", ]
 
+[tool.pyright]
+# used in Pylance extension for language server
+# type check is done by mypy, disable to avoid unwanted errors
+typeCheckingMode = "off"
+
 [tool.isort]
 profile = "black"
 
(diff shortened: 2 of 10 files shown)


NormannK:
thats great!

pyproject.toml
```diff
@@ -43,6 +48,5 @@ ignore = [
 
 [tool.pytest.ini_options]
 minversion = "8.3.3"
-addopts = "--cov=src"
```
noootch, line 46: I had to remove this as when this option is set by default the debugger does not work properly.

noootch: > Option should be added to Makefile and .github/workflows/pytest.yml instead. pytest.yml is using wrongly modules instead of src.

I already fixed the call in the _pytest.yml_. But what would I need to add to the makefile?

b0661: I would use the same options as in pytest.yml:

```makefile
# Target to run tests.
test:
	@echo "Running tests..."
        .venv/bin/pytest -vs --cov src --cov-report term-missing
```

noootch: Done2

noootch:
also ready to be checked

noootch:
Can this PR be merged? It's so cool to be able to debug like this:
grafik

drbacke:
Please re-run all tests

.vscode/extensions.json
NormannK, line 1: comment in json files? afaik thats not allowed

noootch: vscode uses comments within their files, too
grafik

Also the comments will be removed automatically by _**swellaby.workspace-config-plus**_

NormannK: ok nice. if it works it works. Not having comments i always hated in json.
Quite from Wikipedia:
JSON has seen [ad hoc](https://en.wikipedia.org/wiki/Ad_hoc) usage as a [configuration language](https://en.wikipedia.org/wiki/Configuration_file). However, it does not support [comments](https://en.wikipedia.org/wiki/Comment_(computer_programming)). In 2012, Douglas Crockford, JSON creator, had this to say about comments in JSON when used as a configuration language: "I know that the lack of comments makes some people sad, but it shouldn't. Suppose you are using JSON to keep configuration files, which you would like to annotate. Go ahead and insert all the comments you like. Then pipe it through JSMin[[33]](https://en.wikipedia.org/wiki/JSON#cite_note-34) before handing it to your JSON parser."[[21]](https://en.wikipedia.org/wiki/JSON#cite_note-DouglasCrockfordComments-22)

PR #138, opened 2024-10-08
Run tests also after merging

michaelosthege:
Previously tests did not run after merging.

This lead to incompatible changes being merged right after one another.

By running tests on main we will notice breaks earlier.

Other options would be to require every PR to rebase on main before merging. This is known as requiring linear history and can be changed in the project settings.

Issue #136, opened 2024-10-08
[BUG]: class_ems nd_array not JSON serializable


drbacke:
### Describe the issue:

1. Conversion ndarray -> lists was removed. Now the Dictionary is not JSON serializable anymore. 
2. None Values are not accepted by JSONIFY 

### Reproduceable code example:

```python
# Prepare output dictionary
        out: Dict[str, Union[np.ndarray, float]] = {
            "Last_Wh_pro_Stunde": last_wh_pro_stunde,
            "Netzeinspeisung_Wh_pro_Stunde": netzeinspeisung_wh_pro_stunde,
            "Netzbezug_Wh_pro_Stunde": netzbezug_wh_pro_stunde,
            "Kosten_Euro_pro_Stunde": kosten_euro_pro_stunde,
            "akku_soc_pro_stunde": akku_soc_pro_stunde,
            "Einnahmen_Euro_pro_Stunde": einnahmen_euro_pro_stunde,
            "Gesamtbilanz_Euro": gesamtkosten_euro,
            "E-Auto_SoC_pro_Stunde": eauto_soc_pro_stunde,
            "Gesamteinnahmen_Euro": np.sum(einnahmen_euro_pro_stunde),
            "Gesamtkosten_Euro": np.sum(kosten_euro_pro_stunde),
            "Verluste_Pro_Stunde": verluste_wh_pro_stunde,
            "Gesamt_Verluste": np.sum(verluste_wh_pro_stunde),
            "Haushaltsgeraet_wh_pro_stunde": haushaltsgeraet_wh_pro_stunde,
        }
```

### Version information:

-

NormannK:
assign it to me

PR #140, opened 2024-10-08
Json serialization fix

NormannK:
- Standardization of conversion from NumPy arrays
- Central conversion at the place of use
- added case NaN to the test
- Renaming the function to_dict() to to_return_dict() to avoid name clashes with the similarly named pandas function

### DIFF
--- a/src/akkudoktoreos/class_optimize.py
+++ b/src/akkudoktoreos/class_optimize.py
@@ -331,7 +331,7 @@ class optimization_problem:
             "discharge_hours_bin": discharge_hours_bin,
             "eautocharge_hours_float": eautocharge_hours_float,
             "result": o,
-            "eauto_obj": ems.eauto.to_dict(),
+            "eauto_obj": ems.eauto.to_return_dict(),
             "start_solution": start_solution,
             "spuelstart": spuelstart_int,
             "simulation_data": o,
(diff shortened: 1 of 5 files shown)


NormannK:
the test does not finish bc of #138 
its a fix for #136

src/akkudoktoreos/class_optimize.py
```diff
@@ -331,7 +331,7 @@ def optimierung_ems(
             "discharge_hours_bin": discharge_hours_bin,
```
noootch, line 331: instead of this dict the structure could be described with a pydantic model which can be dumped to json. It is also capable of converting numpy arrays.

drbacke:
> the test does not finish bc of #138 its a fix for #136

Thanks @NormannK I will check this out and try to fix more things :-)

drbacke:
@NormannK The global_start_hour in Haushaltsgeräte does not work. I also do not understand the concept

NormannK:
> @NormannK Could you please make a git pull before you make a Pull Request, your Code ist many commits behind

I'm sorry I don't understand. this PR is based on my fork which is up to date and in the checks it says " no conflicts with the base branch". Also the diffs are exactly what I intended to change.
I don't see the issue. 

But im also absolutely ok with not merging my PR's if there is a better solution.
Than again this PR did not touch Haushaltsgeräte.

PR #143, opened 2024-10-09
wip: allow easy access to contribution

BerndCzech:
- [x] remove maria-db integration in `docker-compose.yaml` see: https://github.com/ Akkudoktor-EOS/EOS/pull/112
- [x] small fixes i came across while trying to run docker
- [x] opinion: i removed the `.dockerignore` file and set the components necessary to build as `COPY ...` in `Dockerfile` - imho we should prefer opt-in instead of having everyone adding new folders to consider the docker image size


BerndCzech:
I saw many similarities to https://github.com/Akkudoktor-EOS/EOS/pull/129

I would post it again after 129 got merged and tests on main are running again.

PR #148, opened 2024-10-09
Fix test_server Closes #147

### LINKED ISSUE #147: [BUG]: unit Test Server Always fails
### Describe the issue:

BUG]: unit Test Server Always fails

_No response_

### Error message:

_No response_

### Version information:

-

Lasall:
* Don´t rely on calculation per specific day. For now we can assume 48 values (2 days prediction).


tests/test_server.py
```diff
@@ -1,63 +1,14 @@
...
-  0.3391062422788601,
-  0.29391186081171267,
-  0.24686897327498972,
-  0.19620626090788126,
-  0.1570820352949701,
-  0.13332596485058107
-]
-"""
-    )
+    assert len(json.loads(result.text)) == prediction_hours
```
b0661, line 14: You may use
```python
assert len(result.json()) == prediction_hours
```

PR #149, opened 2024-10-09
Replace pytz with stdlib zoneinfo/timezone, Bump python to 3.9

Lasall:
* Timezone support comes with the standard lib since 3.9

Remove unlisted dependency pytz. If we want to keep it (comes as transitive dep atm), then we should put it into requirements.txt.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,7 +7,7 @@ authors = [
 description = "This project provides a comprehensive solution for simulating and optimizing an energy system based on renewable energy sources. With a focus on photovoltaic (PV) systems, battery storage (batteries), load management (consumer requirements), heat pumps, electric vehicles, and consideration of electricity price data, this system enables forecasting and optimization of energy flow and costs over a specified period."
 readme = "README.md"
 license = {file = "LICENSE"}
-requires-python = ">=3.8"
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -7,7 +7,7 @@ authors = [
 description = "This project provides a comprehensive solution for simulating and optimizing an energy system based on renewable energy sources. With a focus on photovoltaic (PV) systems, battery storage (batteries), load management (consumer requirements), heat pumps, electric vehicles, and consideration of electricity price data, this system enables forecasting and optimization of energy flow and costs over a specified period."
 readme = "README.md"
 license = {file = "LICENSE"}
-requires-python = ">=3.8"
+requires-python = ">=3.9"
```
michaelosthege, line 10: #127 already required 3.10
```suggestion
requires-python = ">=3.10"
```

michaelosthege (approved):

pyproject.toml
NormannK: the need for 3.10 #127 is not mandatory. if there any reasons against it we can remove that with a 2 line change.

Lasall: As long as that part is in the code, the project requires 3.10.

PR #157, opened 2024-10-10
Tests for class_load_container

NormannK:
- translate class_load_container.py
- add type hints
- removed Numpy (we converted it right away after generating back)
- created test for class

should have been a 2nd PR:
Renamed the to_dict() function to return_dict() to avoid conflict with Pandas' to_dict() method and to prevent confusion with similarly named functions like to_list().
(i learned from it)


drbacke:
@NormannK Could you rebase / merge the conflicts please? Then I can test it.

src/akkudoktoreos/class_akku.py
```diff
@@ -31,7 +31,7 @@ def __init__(
         self.min_soc_wh = (self.min_soc_prozent / 100) * self.kapazitaet_wh
         self.max_soc_wh = (self.max_soc_prozent / 100) * self.kapazitaet_wh
 
-    def to_dict(self):
+    def return_dict(self):
```
fortrieb, line 34: We should stick to methods names used in Python std libs (`to_dict`).

src/akkudoktoreosserver/flask_server.py
```diff
@@ -165,9 +164,9 @@ def flask_gesamtlast_simple():
         # leistung_wp = wp.simulate_24h(temperature_forecast)  # Simulate heat pump load for 24 hours
         # gesamtlast.hinzufuegen("Heatpump", leistung_wp)  # Add heat pump load to total load calculation
 
-        last = gesamtlast.gesamtlast_berechnen()  # Calculate total load
-        print(last)  # Output total load
-        return jsonify(last.tolist())  # Return total load as JSON
+        last = gesamtlast.calculate_total_load()  # Calculate total load
+        print(last)
```
fortrieb, line 168: Using `logging.debug` instead of `print`?

fortrieb (changes requested):

src/akkudoktoreosserver/flask_server.py
NormannK: sure but my intention was not to clean up the flask_server for now. Lets make that a new PR

drbacke:
Please re-run all tests

NormannK:
done. no changes needed. please some feedback on the change request

src/akkudoktoreos/class_akku.py
drbacke: yes please, return is uncommon

src/akkudoktoreos/class_load_container.py
```diff
@@ -1,39 +1,47 @@
...
-        if not self.lasten:
-            return []
+        if not self.loads:
+            return []  # Return empty list if no loads are present
 
-        # Assumption: All load arrays have the same length
-        stunden = len(next(iter(self.lasten.values())))
-        gesamtlast_array = [0] * stunden
+        # Initialize the total load array with zeros
+        LoadAggregator_array = [0.0] * self.prediction_hours
```
drbacke, line 40: I'm not sure if this is sensfull. The Load Container is now dependend on prediction_hours. I think it's not until now.

src/akkudoktoreos/class_load_container.py
```diff
@@ -1,39 +1,47 @@
...
 
-        for last_array in self.lasten.values():
-            gesamtlast_array = [
-                gesamtlast + stundenlast
-                for gesamtlast, stundenlast in zip(gesamtlast_array, last_array)
-            ]
+        # Sum loads from all sources
+        for last_array in self.loads.values():
+            for i in range(self.prediction_hours):
```
drbacke, line 44: I'm not sure if this is sensfull. The Load Container is now dependend on prediction_hours. I think it's not until now.

drbacke (changes requested):
Some small changes requested.

src/akkudoktoreos/class_akku.py
NormannK: removed

src/akkudoktoreos/class_load_container.py
NormannK: the add_load() function in only called in the flask_server. two times to add the household load to total load calculation.
in the first case its using `future_predictions = adjuster.predict_next_hours(prediction_hours) ` as a data source which is based on `prediction_hours`. So my assumption was to make sure we add always the same amount. But i do get the point. We might want to use the function to add only some parts to the data and we don't want to add all. in that case we would have to create an array which only have the data we want to add. I guess thats not needed. ill revert it and make a small change to how the amount of hours is calculated so its imo better to read.

src/akkudoktoreos/class_load_container.py
NormannK: removed prediction_hours hers. also used numpy again to make it fast.

NormannK:
@drbacke changed as requested. see change request comments for details.

PR #162, opened 2024-10-10
test_load_corrector

NormannK:
- added a test for class_load_corrector.py
- added type hints and comments to the class_load_corrector.py


drbacke:
@NormannK Please resolve the confilicts Thanks

drbacke:
Please re-run all tests

src/akkudoktoreos/class_load_corrector.py
```diff
@@ -1,33 +1,73 @@
...
+        Returns:
+            pd.DataFrame: The merged dataset.
+        """
+        # Convert time columns to datetime in both datasets
         self.predicted_data["time"] = pd.to_datetime(self.predicted_data["time"])
         self.measured_data["time"] = pd.to_datetime(self.measured_data["time"])
 
-        # Ensure both time columns have the same timezone
+        # Localize time to UTC and then convert to Berlin time
+        # Localize time to UTC and then convert to Berlin time
```
michaelosthege, line 70:
```suggestion
        # Localize time to UTC and then convert to Berlin time
```

src/akkudoktoreos/class_load_corrector.py
```diff
@@ -36,19 +76,33 @@ def _merge_data(self):
         )
         self.measured_data["time"] = self.measured_data["time"].dt.tz_convert("Europe/Berlin")
 
-        # Optionally: Remove timezone information if only working locally
+        # Remove timezone information (optional for local work)
+        # Remove timezone information (optional for local work)
```
michaelosthege, line 80:
```suggestion
        # Remove timezone information (optional for local work)
```

src/akkudoktoreos/class_load_corrector.py
```diff
@@ -57,49 +111,90 @@ def calculate_weighted_mean(self, train_period_weeks=9, test_period_weeks=1):
             test_start_date + pd.Timedelta(weeks=test_period_weeks) - pd.Timedelta(hours=1)
         )
 
+        # Split merged data into training and testing datasets
+        # Split merged data into training and testing datasets
```
michaelosthege, line 115:
```suggestion
        # Split merged data into training and testing datasets
```

src/akkudoktoreos/class_load_corrector.py
```diff
@@ -57,49 +111,90 @@ def calculate_weighted_mean(self, train_period_weeks=9, test_period_weeks=1):
...
         self.test_data = self.merged_data[
             (self.merged_data["time"] >= test_start_date)
             & (self.merged_data["time"] <= test_end_date)
         ]
 
+        # Calculate the difference between actual ('Last') and predicted ('Last Pred')
         self.train_data["Difference"] = self.train_data["Last"] - self.train_data["Last Pred"]
 
+        # Separate training data into weekdays and weekends
+        # Separate training data into weekdays and weekends
```
michaelosthege, line 129:
```suggestion
        # Separate training data into weekdays and weekends
```

src/akkudoktoreos/class_load_corrector.py
```diff
@@ -57,49 +111,90 @@ def calculate_weighted_mean(self, train_period_weeks=9, test_period_weeks=1):
...
         weekdays_train_data = self.train_data[self.train_data["DayOfWeek"] < 5]
         weekends_train_data = self.train_data[self.train_data["DayOfWeek"] >= 5]
 
+        # Calculate weighted mean differences for both weekdays and weekends
+        # Calculate weighted mean differences for both weekdays and weekends
```
michaelosthege, line 134:
```suggestion
        # Calculate weighted mean differences for both weekdays and weekends
```

src/akkudoktoreos/class_load_corrector.py
```diff
@@ -123,76 +218,63 @@ def _plot_data(self, data, data_type):
...
-    def evaluate_model(self):
+    def evaluate_model(self) -> None:
+        """
+        Evaluate the model performance using Mean Squared Error and R-squared metrics.
+        """
+        # Calculate Mean Squared Error and R-squared for the adjusted predictions
         mse = mean_squared_error(self.test_data["Last"], self.test_data["Adjusted Pred"])
         r2 = r2_score(self.test_data["Last"], self.test_data["Adjusted Pred"])
         print(f"Mean Squared Error: {mse}")
         print(f"R-squared: {r2}")
```
michaelosthege, line 229:
```suggestion
    def evaluate_model(self) -> Tuple[float, float]:
        """
        Evaluate the model performance using Mean Squared Error and R-squared metrics.
        
        Args:
            mse: Mean squared error of the adjusted prediction w.r.t. last test data.
            r2: R2 score of the adjusted prediction w.r.t. last test data.
        """
        # Calculate Mean Squared Error and R-squared for the adjusted predictions
        mse = mean_squared_error(self.test_data["Last"], self.test_data["Adjusted Pred"])
        r2 = r2_score(self.test_data["Last"], self.test_data["Adjusted Pred"])
        print(f"Mean Squared Error: {mse}")
        print(f"R-squared: {r2}")
        return mse, r2
```

src/akkudoktoreos/class_load_corrector.py
```diff
@@ -1,33 +1,73 @@
+from typing import Optional
```
michaelosthege, line 1:
```suggestion
from typing import Optional, Tuple
```

src/akkudoktoreos/class_load_corrector.py
```diff
@@ -123,76 +218,63 @@ def _plot_data(self, data, data_type):
...
+        # Get the latest time in the merged data
         last_date = self.merged_data["time"].max()
+
+        # Generate future timestamps for the next 'hours_ahead'
         future_dates = [last_date + pd.Timedelta(hours=i) for i in range(1, hours_ahead + 1)]
         future_df = pd.DataFrame({"time": future_dates})
+
+        # Extract hour and day of the week for the future predictions
+
+        # Extract hour and day of the week for the future predictions
```
michaelosthege, line 250:
```suggestion
        # Extract hour and day of the week for the future predictions
```

src/akkudoktoreos/class_load_corrector.py
```diff
@@ -123,76 +218,63 @@ def _plot_data(self, data, data_type):
...
         future_df["Hour"] = future_df["time"].dt.hour
         future_df["DayOfWeek"] = future_df["time"].dt.dayofweek
+
+        # Predict the load and apply adjustments for future predictions
+
+        # Predict the load and apply adjustments for future predictions
```
michaelosthege, line 256:
```suggestion
        # Predict the load and apply adjustments for future predictions
```

src/akkudoktoreos/class_load_corrector.py
```diff
@@ -123,76 +218,63 @@ def _plot_data(self, data, data_type):
...
-#     selected_columns['time'] = pd.to_datetime(selected_columns['timestamp']).dt.floor('H')
-#     selected_columns['Last'] = pd.to_numeric(selected_columns['Last'], errors='coerce')
-
-#     # Drop rows with NaN values
-#     cleaned_data = selected_columns.dropna()
 
-#     print(cleaned_data)
-#     # Create an instance of LoadForecast
-#     lf = LoadForecast(filepath=r'.\load_profiles.npz', year_energy=6000*1000)
+        # Return forecast for the specific hour, or NaN if hour is out of range
```
michaelosthege, line 277:
```suggestion
```

tests/test_load_corrector.py
```diff
@@ -0,0 +1,187 @@
...
+    adjuster.merged_data = adjuster._merge_data()
+
+    # Calculate the weighted mean over training and testing periods
+    adjuster.calculate_weighted_mean(train_period_weeks=1, test_period_weeks=1)
+
+    # Assert that weekday and weekend differences are calculated and non-empty
+    assert adjuster.weekday_diff is not None, "Weekday differences should be calculated"
+    assert adjuster.weekend_diff is not None, "Weekend differences should be calculated"
+    assert len(adjuster.weekday_diff) > 0, "Weekday differences should not be empty"
+    assert len(adjuster.weekend_diff) > 0, "Weekend differences should not be empty"
```
michaelosthege, line 123:
```suggestion
    assert adjuster.weekday_diff is not None, "Weekday differences should be calculated"
    assert len(adjuster.weekend_diff) > 0, "Weekend differences should not be empty"
```

tests/test_load_corrector.py
```diff
@@ -0,0 +1,187 @@
...
+    adjuster.calculate_weighted_mean(train_period_weeks=1, test_period_weeks=1)
+    adjuster.adjust_predictions()
+
+    # Assert that the 'Adjusted Pred' column is present in both train and test data
+    assert (
+        "Adjusted Pred" in adjuster.train_data.columns
+    ), "Train data should have 'Adjusted Pred' column"
+    assert (
+        "Adjusted Pred" in adjuster.test_data.columns
+    ), "Test data should have 'Adjusted Pred' column"
```
michaelosthege, line 146:
```suggestion
    assert (
        "Adjusted Pred" in adjuster.train_data.columns
    ), "Train data should have 'Adjusted Pred' column"
```

tests/test_load_corrector.py
```diff
@@ -0,0 +1,187 @@
...
+    Test the evaluate_model method to ensure it prints evaluation metrics (MSE and R-squared).
+    """
+    measured_data, predicted_data, load_forecast = setup_data
+    adjuster = LoadPredictionAdjuster(measured_data, predicted_data, load_forecast)
+    adjuster.merged_data = adjuster._merge_data()
+
+    # Calculate weighted mean, adjust predictions, and evaluate the model
+    adjuster.calculate_weighted_mean(train_period_weeks=1, test_period_weeks=1)
+    adjuster.adjust_predictions()
+    adjuster.evaluate_model()
```
michaelosthege, line 163: With the suggested returns above this can be simplified

```suggestion
    mse, r2 = adjuster.evaluate_model()
    assert not np.isnan(mse)
    assert not np.isnan(r2)
```

michaelosthege (commented):
Testing is good, but we should be careful to avoid locking down the codebase to the current architecture and making it harder to do important refactoring in the future.

For example, here are two things that if addressed could make things easier in the long run:

1. The `LoadPredictionAdjuster` is extremely stateful. The relevant data are stored as attributes that are set or modified by calling the methods in exactly the correct order. A scikit-learn pipeline might be more suited for this sequence of data transformations.
2. The `LoadPredictionAdjuster` is initialized with this magic `load_forecast: object`, but it doesn't actually need the `load_forecast`! All it needs is a callable with the same signature as `get_daily_stats` which [can be type annotated as a `Callable`](https://docs.python.org/3/library/typing.html#annotating-callable-objects). Then the test also wouldn't need a `MagicMock`.

NormannK:
thanks for you thorough review. I don't know what happen with those double lines. 
Should we address your remarks for the long run in this PR or produce a milestone/issue for this?

michaelosthege:
> thanks for you thorough review. I don't know what happen with those double lines. Should we address your remarks for the long run in this PR or produce a milestone/issue for this?

Maybe address 2. in this PR because it simplifies the test?

I will propose 1. in a discussion together with some other proposals along the same line.

PR #141, opened 2024-10-08
Json configuration

noootch:
Add a json configuration.

To discuss/questions:

1. Should the mariadb config stay in env?
2. Shall the config.json path be provided as arg or as env var so that it is not saved within the package?

# TODO
- [x] add test
- [x] discuss

PR #163, opened 2024-10-10
Migrate from Flask to FastAPI

Lasall:
* See #119

I tried to keep the changes (almost) minimal. The tests are passing but manual testing is required. To avoid unnecessary errors, I changed the input names to the constructor arguments and moved most of them into a pydantic model class.
In some cases (PVAkku, EAutoAkku), different defaults were used, so I created inherited models, not sure if thats the pydantic way.

FastAPI provides input validation (pydantic) and generates a swagger interface automatically when starting the server. ~~The API should probably be exported to github pages or so (e.g. openapi specs).~~

### Todo:
- [ ] someone with knowledge of the system has to validate its functionality
- [x] update single_test_optimization.py
- [x] pydantic in requirements.txt and decision if version 1 or 2 (currently 2)
- [x] API is linked at swagger.io with spec from this repo and added to spynx


NormannK:
this looks like a lot of work. I'm currently producing tests for the classes we don't have any yet. 
Do you think we should wait for this to push until we have all tests?

On the other hands the already successful tests do look very good.

Lasall:
More tests make it probably safer.

Lasall:
Could fix another bug with the help of single_test_optimization.py. We should probably add more calculation tests. Some sample requests with old and new server result in the same output (with the pdf it's fortunately easy to check) so from my side this can be thoroughly tested now.

NormannK:
I thought about hashing the pdf and checking it vs the new version. what do you think?

Lasall:
Since the pdf is generated from the optimize output, it would be enough to check that step imho.

drbacke:
Thanks, i will check that in my local system 👍 
I will need some days, because i Work on ac Charge, which is wuite a Lot to change

Lasall:
Moved the README info directly into the code (will be exported with Swagger API-doc). I noticed the `simulation_data` that is just a copy of `result`. Do we want to fill that in the future or can I remove it? (atm I disabled that code)

tests/test_openapi.py
```diff
@@ -0,0 +1,20 @@
...
+DIR_PROJECT_ROOT = Path(__file__).parent.parent
+DIR_TESTDATA = Path(__file__).parent / "testdata"
+
+
+def test_openapi_spec_current():
+    """Verify the openapi spec hasn´t changed."""
+
+    old_spec_path = DIR_PROJECT_ROOT / "openapi.json"
+    new_spec_path = DIR_TESTDATA / "openapi-new.json"
+    generate_openapi(new_spec_path)
```
b0661, line 15: Could you use formatted json for openapi(-new).json?

I changed a comment in the source that ended up in the openapi-new.json. It is very hard to find the culprit in a one line diff that basically tells you the whole line changed.

Lasall: Sorry about that, spec is now pretty printed.

drbacke:

NormannK:
personally id like to push the PR #141 first. im currently looking into a failed test in main. after that i might be able to merge 141

NormannK:
@Lasall done

Lasall:
Some endpoints (/strompreis) are broken (same behavior on main) or are missing complete pydantic API information (e.g. /gesamtlast). Tests and /optimize, /pdf endpoints seem to be ok.

NormannK:
its bad we don't have tests for them. what do you think? Migrate to Flask or fix the bugs first?

openapi.json
```diff
@@ -0,0 +1,940 @@
+{
```
b0661, line 1: You may use https://github.com/Akkudoktor-EOS/EOS/blob/main/docs/akkudoktoreosserver/openapi.json for testing. This way the documentation will stay in sync with the server api.

Lasall:
I moved the server entrypoint to akkudoktoreos. However I still fully removed flask as we can probably use the archived file in git for reference. The logic within the endpoints was mostly untouched besides data validation.

README.md
```diff
@@ -111,250 +119,7 @@ These classes work together to enable a detailed simulation and optimization of
...
-- **Type**: null or object
-
-### start_solution
-
-- **Description**: An array of binary values (0 or 1) representing a possible starting solution for the simulation.
-- **Type**: Array
-- **Element Type**: Integer (0 or 1)
-- **Length**: 48
+See the Swagger documentation for detailed information: [EOS OpenAPI Spec](https://petstore3.swagger.io/?url=https://raw.githubusercontent.com/Akkudoktor-EOS/EOS/refs/heads/main/openapi.json)
+>>>>>>> 4b4cbf2 (Move API doc from README to pydantic model classes (swagger))
```
b0661, line 125: Seems to be a leftover.

NormannK:
@Lasall @b0661 as soon as you both its ready to merge ill click the button.

b0661:
@NormannK That’s fine with me.

Lasall:
Ready to merge for me too. I looked over the MR and compared the flask side-by-side as the diff is not so good (e.g. different indentation). Found one smaller merge mistake (optimization test data generation), see last commit.

NormannK:
Ill merge it this evening. I made PR'S for the dead endpoints. Ill adjust it after

NormannK (approved):

PR #164, opened 2024-10-11
inverter class rewritten

NormannK:
fully rewritten the inverter class
add tests

please check carefully
output from single_test_optimization.py is identical


drbacke:
We should BE very carefull Here and a Lot of more Tests are required. The Inverter behaviour is very sensitive to Changes and Bugs are hard to find

NormannK:
I increased the test coverage we do now test for:

1. Excess Generation: The system handles more generation than consumption, feeding excess to the grid and charging the battery.
2. Generation Equals Consumption: No feed-in or grid draw when generation matches consumption.
3. Battery Discharge: The battery discharges to help meet consumption, and the grid supplies any remaining shortfall.
4. Battery Empty: When the battery is empty, the grid covers the entire shortfall.
5. Battery Full at Start: When the battery is full, excess generation is fed into the grid.
6. Insufficient Generation, No Battery: When there’s insufficient generation and no battery discharge, the grid supplies the entire shortfall.
7. Insufficient Generation with Battery Assistance: The battery partially discharges to cover a shortfall, with the grid supplying the rest.
8. Zero Generation: When there’s zero generation, the system relies entirely on the battery and grid.

let me know if you have other scenarios we should test.

NormannK:
There are some things that should be done too but would extend the logic and im not sure how other parts of the optimization would react.
I would not allow it to handle wrong values.
e.g. you can hand over a negative generation value. This cant handle and something should produce an error or do something. e,g, adding:
```
        if generation < 0:
            generation = 0.0  # Treat invalid negative generation as zero
```

NormannK:
this should be the best test. i pack ported the new test for the new class to the old class by only changing the import and the names of the function/variables called. minimal changes. 
both files with _old should be deleted in a later PR if accepted.

drbacke:

NormannK:
done. only a single very easy conflict and the new test did need the new class

NormannK:
redo tests

NormannK:
@drbacke please review this PR. I did every safety check i can think of. Before merging both _old files should be removed.

src/akkudoktoreos/class_inverter.py
```diff
@@ -1,58 +1,49 @@
...
+class Inverter:
+    def __init__(self, max_power_wh: float, battery) -> None:
+        self.max_power_wh = max_power_wh  # Maximum power inverter can handle
+        self.battery = battery  # Battery object
+
+    def process_energy(
+        self, generation: float, consumption: float, hour: int
+    ) -> Tuple[float, float, float, float]:
+        losses = 0.0
+        grid_feed_in = 0.0
```
SREbuilt, line 13: rename grid_feed_in --> grid_export and
rename grid_draw    --> grid_import

PR #165, opened 2024-10-11
Make config more flexible using env vars

niklasbeinghaus:
* Adds tests for config.py
* Adds support for environment variables in config.py

### DIFF
--- a/src/akkudoktoreos/class_optimize.py
+++ b/src/akkudoktoreos/class_optimize.py
@@ -8,7 +8,7 @@ from akkudoktoreos.class_akku import PVAkku
 from akkudoktoreos.class_ems import EnergieManagementSystem
 from akkudoktoreos.class_haushaltsgeraet import Haushaltsgeraet
 from akkudoktoreos.class_inverter import Wechselrichter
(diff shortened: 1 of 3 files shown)


NormannK:
i would not test for specific values in the config. the config is meant to be edited imo. 
How about checking only for the presence of the option and maybe sometimes the limits of the value.

Unless we argue this is only meant for the current dev phase and we should not touch those values

Lasall:
Fyi: There is an extensive config overhaul here: https://github.com/Akkudoktor-EOS/EOS/pull/141 should probably put the env support on top of that.

noootch:
> Fyi: There is an extensive config overhaul here: #141 should probably put the env support on top of that.

Yes, I'll push my final solution in a moment. And it will have env support : )

noootch:
Ready for review :D https://github.com/Akkudoktor-EOS/EOS/pull/141

NormannK:
all tests are working. single_test_optimization.py output is perfect. pdf looks as it should be. although we are missing the last page since some PR's

drbacke:

NormannK:
outdated, we went for json #141

PR #171, opened 2024-10-13
Add more optimization tests

Lasall:
Add more optimization tests.
    * Include test run from single_test_optimization into pytest run.
    * Check for complete result with pytest.approx (1e-6 tolerance) to spot
       changes to expected result quickly.
    * Add one long test run (skipped by default).
    * Use --full-run pytest option to activate long test run (400 gen like
       single_test_optimization).

 Test PDF   
     * Mock pdf creation away in optimize test.
     * Add new PDF creation test (test by image comparison).

I wonder if the tolerance is enough or we should increase it. I'm not sure how much the solutions can divert. With the full run (copy from single_test_optimization) it probably makes sense to keep it narrow.

A comment in the code mentioned to test the optimization in unit tests, however until those might or might not be in place, we should have some sort of verification of the status-quo.


NormannK:
thank you very much for this much needed test.
during my tests i have never seen the need for a 1e-6 tolerance but changes that low should not be a problem either.
do you think we should use a Git Pre-Commit or Pre-Push Hook to run the full test once there is a file change in the class_optimize, ems, PVAkku or Inverter?

drbacke:

tests/test_class_optimize.py
```diff
@@ -1,16 +1,39 @@
 import json
 from pathlib import Path
+from typing import Any
 
 import pytest
+from mock import patch
```
michaelosthege, line 6:
```suggestion
from typing import Any
from unittest.mock import patch

import pytest
```

requirements-dev.txt
```diff
@@ -1,4 +1,5 @@
 -r requirements.txt
+mock==5.1.0
```
michaelosthege, line 2:
```suggestion
```

michaelosthege (commented):
mock became part of the `unittest` standard library module in Python 3.3, so we don't need to add a dependency.

Otherwise this looks good 👍

PR #167, opened 2024-10-11
Cleanup: Fix violin chart labels, remove debug code

Lasall:

PR #178, opened 2024-10-24
Visualize reworked

NormannK:
new structured class to visualize the results in a pdf, dividing plot generation from data ingestion.
see example usage in the class file.

- usage of templates for each graph type
- clean, automated page layout
- added in parallel to the class_optimisation to not break existing tests and ongoing PR's e.g.  #167 and #171 
- violin chart is intentionally missing until #167 is accepted (also i have trouble understanding it)


NormannK:
[grouped_energy_report.pdf](https://github.com/user-attachments/files/17499421/grouped_energy_report.pdf)
[example_report.pdf](https://github.com/user-attachments/files/17499432/example_report.pdf)

first pdf is the result from the single_test_optimization. The 2nd is the example result from the class.
I'd be happy about feedback. I'll keep this in a draft until those other PR's are done.

NormannK:
full functional now. ready for review. all comments before updated.
class_optimize is now returning the extra_data. tests updated to see that.

NormannK:
#203 was not needed here.

NormannK:
@Lasall could you please help me fixing the tests? Im struggling hard to get it working.

Lasall:
You are replacing the visualisiere_ergebnisse with prepare_visualize? In any case in the optimization tests, the visualization should be excluded, so you can just add another patch after the @patch for visualisiere_ergebnisse for the prepare_visualize (and add it as first parameter to the test_optimize function) and then later just check if the function was called.

Lasall:
With the visualization test, the test will take the dict input data you get from the optimierung_ems result. To get the data directly from the test you can comment in the json.dump lines in the test_class_optimize.py and take the json from there as new input data. Since you now use the parameters as well you probably have to export them and and merge into a big json ({parameters: ...., results: ...}) then later parse them separately: Load the json input data as json into dict as it's already done in the visualization test. Then convert the parameters data from dict to parameters: OptimizationParameters(**loaded_data["parameters"]) and access the results data from the loaded json: loaded_json["results"].

If my explanation is too cryptic I can generate that data for you. Btw. the visualization test atm only compares the first page of the pdf (I didn't want to add any pdf utility dependency).

src/akkudoktoreos/class_optimize.py
```diff
@@ -645,5 +645,8 @@ def optimierung_ems(
             "eauto_obj": ems.eauto.to_dict(),
             "start_solution": start_solution,
             "spuelstart": spuelstart_int,
-            # "simulation_data": o,
+            "simulation_data": o,
+            "extra_data": extra_data,
```
Lasall, line 649: If the API changes, please extend OptimizeResponse as well (extra_data). Do we need simulation_data? Atm it's just a copy of results.

tests/test_class_optimize.py
```diff
@@ -64,10 +64,11 @@ def test_optimize(
 
     # Call the optimization function
     ergebnis = opt_class.optimierung_ems(parameters=input_data, start_hour=start_hour, ngen=ngen)
+    # from akkudoktoreos.class_numpy_encoder import NumpyEncoder
+
     # with open(f"new_{fn_out}", "w") as f_out:
-    #     from akkudoktoreos.class_numpy_encoder import NumpyEncoder
```
Lasall, line 68: probably a merge/rebase artifact, however here you probably want to export the parameters as well so they can be used as visualization test input.

src/akkudoktoreos/class_optimize.py
NormannK: you are right, no we do not.

Issue #183, opened 2024-10-27
[BUG]: class_akku.py min_soc_percent may not be Null


akrauss:
### Describe the issue:

I am currently testing the EOS with Home Assistant. Unfortunately I'm quite a noob when it comes to programming, but I can test ;)

In the class_akku.py the default value for min_soc_percent is Null, which leads to an error in the calculation of min_soc_wh.

Default value should be 0 not zero ....

### Reproduceable code example:

```python
src/akkudoktoreosserver/flask_server.py

        # Optional min SoC PV Battery
        if "min_soc_prozent" not in parameter:
            parameter["min_soc_prozent"] = None
```

### Error message:

```shell
eos-1  | Traceback (most recent call last):
eos-1  |   File "/usr/local/lib/python3.12/site-packages/flask/app.py", line 1498, in __call__
eos-1  |     return self.wsgi_app(environ, start_response)
eos-1  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
eos-1  |   File "/usr/local/lib/python3.12/site-packages/flask/app.py", line 1476, in wsgi_app
eos-1  |     response = self.handle_exception(e)
eos-1  |                ^^^^^^^^^^^^^^^^^^^^^^^^
eos-1  |   File "/usr/local/lib/python3.12/site-packages/flask/app.py", line 1473, in wsgi_app
eos-1  |     response = self.full_dispatch_request()
eos-1  |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
eos-1  |   File "/usr/local/lib/python3.12/site-packages/flask/app.py", line 882, in full_dispatch_request
eos-1  |     rv = self.handle_user_exception(e)
eos-1  |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
eos-1  |   File "/usr/local/lib/python3.12/site-packages/flask/app.py", line 880, in full_dispatch_request
eos-1  |     rv = self.dispatch_request()
eos-1  |          ^^^^^^^^^^^^^^^^^^^^^^^
eos-1  |   File "/usr/local/lib/python3.12/site-packages/flask/app.py", line 865, in dispatch_request
eos-1  |     return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)  # type: ignore[no-any-return]
[...]

PR #184, opened 2024-10-27
set min_soc_prozent 0 by default

NormannK:
fix for #183. The current default value of "Null" is not compatible with the class_akku. A default value of 0 is more reasonable.
We already set the value to 0 as default when we initialize the PVAkku class.


NormannK:
I don't understand why the test for test_class_optimize.py is failing. The change does not touch that. Please someone take a look.

Lasall:
Probably some bug with the dailight saving. Same error on main.

NormannK:
good call. Time for a new bug report.

PR #188, opened 2024-11-01
Structure code in logically separated submodules

michaelosthege:
I went through the codebase and tried to restructure things in logically separated submodules.

The dependencies are as follows: `devices` ← `prediction` ← `optimization`

More background on why I think this should be done: https://github.com/Akkudoktor-EOS/EOS/discussions/187


NormannK:
i love it, but this will break all current PR'S :-D

b0661:
Looks very promising. I'm definitivly for creating submodules. I would propose to also add a utilities submodules for helpers like datetime conversion  and cache file handling ... Maybe config should also get it's own submodule.

SREbuilt:
> i love it, but this will break all current PR'S :-D

major refactoring is always painful, and in case the benefits out weight the pain, then let's go that route. This refactoring will pay off in my opinion (as you did).
Here the benefits will be that each submodule can be developed (to some degree) individually from any other... having robust interfaces is a prerequisite. Any interfaces must be kept stable for long-term.
Shall we align in a meeting and agree on the interface definition as a next step?
what do others think should be a next step?

SREbuilt (commented):
looks all good to me 🙏
I like the abstraction of devices a lot.

NormannK:
I would merge this next in line

b0661 (changes requested):
Could utils please become a submodule not just a python file. There are already several utilities and this would really help to organize them.

NormannK:
> Could utils please become a submodule not just a python file. There are already several utilities and this would really help to organize them.
im good with either solution. but as a 3rd way we could get rid of it.
NumpyEncoder can be done with build in tools
ist_dst_wechsel did not work the last time iirc and also we wanted to change the time handling in a general way.

src/akkudoktoreos/optimization/problem.py
NormannK, line 1: im not happy with that file name. i understand how you got there, but there is no problem in this file. there are only solutions coming out :-)
how about we call it "solver.py" or something along that. or solver_genetic.py to be ready to use other types.

b0661:
> > Could utils please become a submodule not just a python file. There are already several utilities and this would really help to organize them.
> 
> im good with either solution. but as a 3rd way we could get rid of it.

We have cachefilestore.py, datetimeutil.py, logutil.py already. These were created by me and should go to utils, and you shouldn't get rid of them !-). 

I don't know about the NumpyEncoder. A daylight saving time check can be added to datetimeutil.py.

NormannK:
> > > Could utils please become a submodule not just a python file. There are already several utilities and this would really help to organize them.
> > 
> > 
> > im good with either solution. but as a 3rd way we could get rid of it.
> 
> We have cachefilestore.py, datetimeutil.py, logutil.py already. These were created by me and should go to utils, and you shouldn't get rid of them !-).
> 
> I don't know about the NumpyEncoder. A daylight saving time check can be added to datetimeutil.py.

you are right. yeah lets go with your suggestion

michaelosthege:
I strongly disagree: EOS is not a package of utilities functions!

Instead of moving these utility modules into a submodule, I would suggest to replace them by well-maintained external dependencies that EOS maintainers won't have to worry about.

* `logutil.py`: just use `logger = logging.getLogger(__name__)` like every other Python project does. Don't reinvent the wheel!
* `datetimeutil.py`: just use the existing, extremely popular [`dateutil`](https://github.com/dateutil/dateutil) library if the standard library isn't good enough. Or __raise errors when someone provides non iso-formatted dates!__
* `cachefilestore.py`: 360 lines just to cache something with a timeout? [`cachetools`](https://github.com/tkem/cachetools) already does that. And unless EOS crashes every hour, it should be totally fine to just cache in-memory.

NormannK:
I propose that we accept this PR for now and address the identified issues through smaller, more focused, and easily reviewable follow-up PRs.

Currently, we're discussing two topics in parallel:
1.Where to store the elements we have and what naming conventions to use.
2. What we need and how we might replace it.

This PR primarily addresses the first topic.

That said, we should aim to use built-in or well-established functions wherever possible. I'm not very familiar with the [dateutil](https://github.com/dateutil/dateutil) library or the caching mechanisms involved, so any insights or recommendations would be greatly appreciated.

It's worth noting that most of this util functions stems from PR #174. However, nothing is set in stone, so we can refine the classes further once we have a clearer understanding of the data structure.

b0661:
> * `logutil.py`: just use `logger = logging.getLogger(__name__)` like every other Python project does. Don't reinvent the wheel!

logutil does some project wide configuration that was centralized and allows to centralize it. That is it is a specialization for the project.

> * `datetimeutil.py`: just use the existing, extremely popular [`dateutil`](https://github.com/dateutil/dateutil) library if the standard library isn't good enough. Or **raise errors when someone provides non iso-formatted dates!**

datetimeutil also specializes the time date handling for the project. It provides functionality not provided by pure datetime or Pendulum. It allows to change the implementation behind the function, e.g. switch from datetime to pendulum. Date conversion functions were distributed all over the code with date formats that need format hints to convert. Timezone handling was not consistent. **raise errors when someone provides non iso-formatted dates!** is a funny concept if the non iso-formatted dates are coming from the source of your prediction.

> * `cachefilestore.py`: 360 lines just to cache something with a timeout? [`cachetools`](https://github.com/tkem/cachetools) already does that. And unless EOS crashes every hour, it should be totally fine to just cache in-memory.

If EOS is fine with caching in memory, just remove it. The original implementation cached to files, I did not want to change the concept, when adding tests to pvforecast. Here it is the same as with the other functions, centralization of functionality that was otherwise distributed all over the code several times.

So all these tools specialize something in a way for EOS.

> Instead of moving these utility modules into a submodule, I would suggest to replace them by well-maintained external dependencies that EOS maintainers won't have to worry about.

I agree. I looked into file caching and also asked you, but there were no good candidates. Maybe you now know or we just skip file caching.

For the other utilities, I think they provide an EOS specific protocol for certain functionality. Mostly they are shims with small extensions around well maintained libraries. None of it is a library as such.

> I strongly disagree: EOS is not a package of utilities functions!

This is a straw man. Nobody want's to create a package of utilities. But as in every project there is some common functionality that should be maintained at a central place. For ease of maintenance and for ease of testing. If you want to call the module 'common` this would be just another name for the the same concept.

So to come to the end of this length discussion.

I strongly disagree to **not** have a place to gather functionality that is common to all EOS and that is not directly linked to one of the functional concepts like optimisation, prediction, ... I don't care about the naming. If 'common` better fits, use it or something else.

michaelosthege:
Since I had to rebase yet again, I moved the utilities into `utils` and renamed `problem.py` → `genetic.py`.

Once this is merged I will open a PR to reduce some complexity.

NormannK:
This PR has been the subject of a lot of discussion for quite some time, with quite a bit of back and forth. I really appreciate everyone taking the time to dive in and share their thoughts. While not every part of this PR will please everyone, I think overall it makes the code easier to work with and better as a whole.

I also hope some of the points brought up here can spark new PRs and keep the conversation going. Thanks again to everyone who contributed!

PR #192, opened 2024-11-11
fix test_timezone_behaviour in windows

NormannK:
The issue here is that time.tzset() is available on Unix-based systems (Linux, macOS) but is not supported on Windows. This function applies changes to the environment variable TZ for Unix systems, but there's no direct equivalent in Windows.

To make the test pass across platforms, we need a solution that avoids time.tzset() when running on Windows.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import platform
 import subprocess
 import sys
 import time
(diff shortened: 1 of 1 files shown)


b0661:
Maybe just exclude the test case on Windows as long as there is no solution that avoids time.tzset() when running on Windows. This way it is obvious that the test is not run.

I can't contribute here, no Windows development environment.

b0661:
Could you please check wether #195 would work for you.

PR #193, opened 2024-11-11
pv_forecast_result_1 last digit rounding change

NormannK:
looks like the last digit rounding did change.
checked on windows and debian.
should this happen more often we need to change the test accuracy. 
would be great of @b0661 could look into this PR. its based on #174


b0661:
> github is producing different results, its getting ugly

I did not change the reporting code and it worked on my machine and on Github (see closed PR #174).

Floating point precision maybe different on Github due to another change or just by chance, e.g. other server. 

Anyway it does not make sense to compare power on Picowatt level. I don't know the intent to not round the values in the report, but my suggestion is to round them to a sensefull mantissa.

I will try to test tonight.

b0661:
Please see #194. PR is passing tests on GitHub.

PR #200, opened 2024-11-15
endpoints hardening

NormannK:
goal is to improve the endpoints to avoid crashes
- default TZ added to pv_forecast
- /gesamtlast_simple is using the current year if non provided
- /visualization_results.pdf is returning 204 No Content if not found
- added tests for those endpoints, only checking of they are responding

### DIFF
--- a/src/akkudoktoreos/class_pv_forecast.py
+++ b/src/akkudoktoreos/class_pv_forecast.py
@@ -259,7 +259,7 @@ class PVForecast:
         else:
             self._forecast_start = None
         self.prediction_hours = prediction_hours
-        self._tz_name = None
+        self._tz_name = "Europe/Berlin"
 
         if self.data or self.filepath or self.url:
             self.process_data(
--- a/src/akkudoktoreosserver/flask_server.py
+++ b/src/akkudoktoreosserver/flask_server.py
@@ -10,7 +10,7 @@ import matplotlib
 matplotlib.use("Agg")
 
 import pandas as pd
-from flask import Flask, jsonify, redirect, request, send_from_directory, url_for
+from flask import Flask, jsonify, redirect, request, send_from_directory, url_for, Response
 
 from akkudoktoreos.class_load import LoadForecast
 from akkudoktoreos.class_load_container import Gesamtlast
@@ -132,9 +132,13 @@ def flask_gesamtlast():
 @app.route("/gesamtlast_simple", methods=["GET"])
 def flask_gesamtlast_simple():
     if request.method == "GET":
-        year_energy = float(
-            request.args.get("year_energy")
+        year_energy = request.args.get(
+            "year_energy"
         )  # Get annual energy value from query parameters
+        if year_energy is None:
+            year_energy = float(datetime.now().year)
+        else:
+            year_energy = float(year_energy)
         date_now, date = get_start_enddate(
             config.eos.prediction_hours, startdate=datetime.now().date()
         )  # Get the current date and prediction end date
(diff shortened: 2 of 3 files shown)


src/akkudoktoreos/class_pv_forecast.py
```diff
@@ -259,7 +259,7 @@ def __init__(
         else:
             self._forecast_start = None
         self.prediction_hours = prediction_hours
-        self._tz_name = None
+        self._tz_name = "Europe/Berlin"
```
SREbuilt, line 262: instead of hard coding the time zone to Germany, would it make better sense to determine the OS configured time zone and use that?
Or at least move the time zone as a parameter to the JSON config?

src/akkudoktoreosserver/flask_server.py
```diff
@@ -132,9 +140,13 @@ def flask_gesamtlast():
 @app.route("/gesamtlast_simple", methods=["GET"])
 def flask_gesamtlast_simple():
```
SREbuilt, line 141: still a bit of DEnglish here..
is flask_totalload_simple() a better name? 
or how about flask_total_load_simple?

maybe this is not the best time to rename now, as it's in alignment with the endpoints' name... 
Once the endpoint will be renamed, this method shall be renamed accordingly.

src/akkudoktoreosserver/flask_server.py
```diff
@@ -132,9 +140,13 @@ def flask_gesamtlast():
...
     if request.method == "GET":
-        year_energy = float(
-            request.args.get("year_energy")
+        year_energy = request.args.get(
```
SREbuilt, line 143: to me, the name year_energy is ambiguous... also comment is misleading. 
Also, what type of energy is it?: consumed, produced, is it the actual measured energy or is it predicted?

Question: what is the endpoint queried for?
If it is the predicted energy demand for a given year, then I propose the name predicted_annual_energy_demand  or if it is the year then "predicted_annual_energy_demand_for_year" would be precise (and frank speaking a bit bulky; but hey, bits are cheap these days).

Please beg with me, in case I'm completely off topic here... I haven't reviewed the entire code yet.

src/akkudoktoreosserver/flask_server.py
```diff
@@ -132,9 +140,13 @@ def flask_gesamtlast():
...
+            "year_energy"
         )  # Get annual energy value from query parameters
```
SREbuilt, line 145: if it is the year, then the following comment will be more precise:
"get year for predicted annual energy demand".

SREbuilt (commented):
I've commented on a few lines. hope this is helpful.

src/akkudoktoreos/class_pv_forecast.py
NormannK: json is obviously the way to go. i had made a future todo list in an edit in the first comment with this. looks like i didn't hit save. I have not yet worked with the new cfg. if its ok for you i would make a later PR and change much more stuff into the cfg and not only this.

src/akkudoktoreosserver/flask_server.py
NormannK: to be honest: i don't know what this is used for. It was dead for months and nobody complained. i thought about removing it completely. Ill look into the variable names and find better.

src/akkudoktoreos/class_pv_forecast.py
b0661: to_timezone in dateimeutils.py will provide you a timezone based on latitude and longitude. We need latitude and longitude anyway for PV forecast. Currently it is hidden in the url provided for forecast request.
I'm working on to make the forecasts more universal.

PR #217, opened 2024-11-22
Mypy

Lasall:
Add Mypy support:
   * Add to pre-commit (currently installs own deps, could maybe changed to poetry venv in the future to reuse environment and don't need duplicated types deps).
   * Add type hints (default mypy config).

In second commit:
   * Add missing annotations.
   * Enforce type hints.

Remarks:
   * Modules not supporting type hints (sklearn, deap, xprocess) and unused code (class_soc_calc.py) are ignored.
   * genetic.py: deap.creator.Individual is hinted as list[int] atm because it inherits from list and is filled with ints.
   * I was a little bit lazy with the numpy arrays and didn't explicitely write the types there.
   * Thanks to the type checking I found a few bugs and wrong hints.

### DIFF
--- a/src/akkudoktoreos/devices/battery.py
+++ b/src/akkudoktoreos/devices/battery.py
@@ -6,7 +6,7 @@ from pydantic import BaseModel, Field, field_validator
 from akkudoktoreos.utils.utils import NumpyEncoder
 
 
-def max_ladeleistung_w_field(default=None):
+def max_ladeleistung_w_field(default: Optional[float] = None) -> Optional[float]:
     return Field(
         default=default,
         gt=0,
@@ -14,7 +14,7 @@ def max_ladeleistung_w_field(default=None):
     )
 
 
-def start_soc_prozent_field(description: str):
+def start_soc_prozent_field(description: str) -> int:
     return Field(default=0, ge=0, le=100, description=description)
 
 
@@ -108,7 +108,7 @@ class PVAkku:
         self.min_soc_wh = (self.min_soc_prozent / 100) * self.kapazitaet_wh
         self.max_soc_wh = (self.max_soc_prozent / 100) * self.kapazitaet_wh
 
-    def to_dict(self):
+    def to_dict(self) -> dict[str, Any]:
         return {
             "kapazitaet_wh": self.kapazitaet_wh,
             "start_soc_prozent": self.start_soc_prozent,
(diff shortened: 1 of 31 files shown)


src/akkudoktoreos/devices/battery.py
```diff
@@ -1,40 +1,45 @@
...
 
-def max_ladeleistung_w_field(default=None):
+
+def max_ladeleistung_w_field(default: Optional[float] = None) -> Optional[float]:
     return Field(
-        default,
+        default=default,
         gt=0,
         description="An integer representing the charging power of the battery in watts.",
     )
```
michaelosthege, line 14: I'm not familiar with pydantic, but shouldn't the return type hint be `Field`?

Lasall: Field is actually a function but it returns a FieldInfo. The type annotation for the Field function however is Any as return type as it is assigned to the pydantic model field (with a user type like Optional[float]). Therefore this type hint cheats a bit in the way similar as it's done with the deap.creator.Individual -> list[int].

NormannK:
id like to merge this next before it got hit by the very large PR #220 
We be great if the conflicts can be resolved

src/akkudoktoreos/visualize.py
```diff
@@ -81,13 +83,14 @@ def visualisiere_ergebnisse(
...
 
         # Temperature forecast
-        plt.subplot(3, 2, 5)
-        plt.title("Temperature Forecast (°C)")
-        plt.plot(hours, temperature, label="Temperature (°C)", marker="x")
-        plt.xlabel("Hour of the Day")
-        plt.ylabel("°C")
-        plt.legend()
-        plt.grid(True)
+        if temperature is not None:
```
NormannK, line 86: genetic.py line 26: temperature is not optional. so it can't be empty, or is there a chance?

tests/test_pv_forecast.py
```diff
@@ -121,7 +121,7 @@ def test_update_ac_power_measurement(pv_forecast_instance, sample_forecast_start
     forecast_start = pv_forecast_instance.get_forecast_start()
     assert forecast_start == sample_forecast_start
 
-    updated = pv_forecast_instance.update_ac_power_measurement(forecast_start, 1000)
+    updated = pv_forecast_instance.update_ac_power_measurement(1000, forecast_start)
```
NormannK, line 124: I don't mind, but is there a reason the order got changed?

Lasall: the updated value is not optional, so had to move to the front.

src/akkudoktoreos/visualize.py
Lasall: it's unused besides the display at optimization, therefore optional

PR #235, opened 2024-12-11
Price Prediction failed, used Normanns fixes for the new code

drbacke:


drbacke:
Fixes of #199  ported to the new code. Will del the old PR.

Issue #216, opened 2024-11-21
[ENH]: Bessere Simulationsergebnisse durch Optimierung der Zustände


celle1234:
### Proposed implementation

```python
Die verschiedenen Zustände für Laden, Entladen, ... werden nur als eine Zahl codiert.

The states are:
0: No action ("idle")
1: Discharge ("discharge")
2-6: AC charging with different power levels ("ac_charge")
7-8: DC charging Dissallowed/allowed ("dc_charge")

Wenn ich den Code richtig verstehe werden bei den Permutationen die Werte per Random gefüllt. Dabei wird Logischerweise das AC Charge bevorzugt und man kommt oft aus den lokalen Minima der Kosten nicht mehr raus.

Zum testen habe ich die Zustände bei mir wie folgt geändert:
0-4 idle
5-9 discharge 
10-14 AC Charge 0.2, 0.4, 0.6, 0.8, 1

Das gibt deutlich bessere Simulationsergebnisse, die nicht sinnlos den Akku lädt, obwohl das nur Geld kostet und nichts bringt.


diff --git a/src/akkudoktoreos/class_optimize.py b/src/akkudoktoreos/class_optimize.py
index bc0b3f5..b062bde 100644
--- a/src/akkudoktoreos/class_optimize.py
+++ b/src/akkudoktoreos/class_optimize.py
@@ -208,7 +208,8 @@ class optimization_problem:

         # Create ac_charge array: Only consider values between 2 and 6 (AC charging power levels), set the rest to 0
         ac_charge = np.where(
-            (discharge_hours_bin >= 2) & (discharge_hours_bin <= 6), discharge_hours_bin - 1, 0
+            #(discharge_hours_bin >= 2) & (discharge_hours_bin <= 6), discharge_hours_bin - 1, 0
+            (discharge_hours_bin >= 10) & (discharge_hours_bin <= 14), discharge_hours_bin - 9, 0
         )
         ac_charge = ac_charge / 5.0  # Normalize AC charge to range between 0 and 1

@@ -217,12 +218,14 @@ class optimization_problem:
         if self.optimize_dc_charge:
             dc_charge = np.where(discharge_hours_bin == 8, 1, 0)
         else:
-            dc_charge = np.ones_like(
-                discharge_hours_bin
-            )  # Set DC charge to 0 if optimization is disabled
+            dc_charge = np.where(discharge_hours_bin <= 100, 1, 0)
+            #dc_charge = np.ones_like(
[...]

PR #242, opened 2024-12-14
optimization states for AC, DC and IDLE now similar probab. Also AC states taken from config. Maybe a single config option for AC and E-Auto States is sensefull.

drbacke:
[ENH]: Bessere Simulationsergebnisse durch Optimierung der Zustände #216
optimization states for AC, DC and IDLE now similar probab. Also AC states taken from config. Maybe a single config option for AC and E-Auto States is sensefull.


drbacke (commented):
Please check

NormannK (approved):
multi_file_distribution_with_calculated_averages_new
very good results after only 1 iteration with only 200 generations!
[visualization_results_old.pdf](https://github.com/user-attachments/files/18140593/visualization_results_old.pdf)
[visualization_results_new.pdf](https://github.com/user-attachments/files/18140594/visualization_results_new.pdf)
overall results are even better than before! we see more batter usage and more agressive charging when the price is low.

PR #245, opened 2024-12-15
Inverter v2

NormannK:
- rewritten/translated the inverter class
- add tests
- same logic as #164 where I did show that the new tests are also working on the old class

Todo:
- [x] rename the class itself
- [x] translate in devices.py


NormannK:
done, ready for review and merging

src/akkudoktoreos/optimization/genetic.py
```diff
@@ -488,10 +488,10 @@ def optimierung_ems(
         )
 
         # Initialize the inverter and energy management system
-        wr = Wechselrichter(parameters.wechselrichter, akku)
+        wr = Inverter(parameters.inverter, akku)
```
Lasall, line 491: variable name could be translated

NormannK: yes, but this is the pr for the inverter file. I cant translate all parts of all files. that would be an endless rabbit hole. 
imo battery must come next.

tests/testdata/optimize_input_1.json
```diff
@@ -1,58 +1,60 @@
 {
-    "ems": {
```
Lasall, line 2: could you keep indentation=4 so it's easier to see what really changed

src/akkudoktoreos/optimization/genetic.py
NormannK: done anyway ;-)

tests/testdata/optimize_input_1.json
NormannK: hmm visual studio did that automatically. any idea how to prevent it?

Lasall (approved):

PR #247, opened 2024-12-15
Tests for class_load_container v2

NormannK:
translate class_load_container.py
renamed file according to class name
added native type hints
created test for class with human readable tests

2nd try after #157


src/akkudoktoreos/prediction/load_aggregator.py
```diff
@@ -0,0 +1,39 @@
...
+        self.loads[name] = list(last_array)
+
+    def calculate_total_load(self) -> list[float]:
+        """Calculates the total load for each hour by summing up the loads from all sources.
+
+        :return: A list representing the total load for each hour.
+                 Returns an empty list if no loads have been added.
+        """
+        if not self.loads:
+            return []  # Return empty list if no loads are present
```
Lasall, line 34: the list comprehension later should already give an empty list on empty loads or does it error?

NormannK: true! removing it.

src/akkudoktoreos/prediction/load_aggregator.py
```diff
@@ -0,0 +1,39 @@
...
+        """Initializes the LoadAggregator object with the number of prediction hours.
+
+        :param prediction_hours: Number of hours to predict (default: 24)
+        """
+        self.loads: defaultdict[str, list[float]] = defaultdict(
+            list
+        )  # Dictionary to hold load arrays for different sources
+        self.prediction_hours: int = prediction_hours
+
+    def add_load(self, name: str, last_array: list[float] | tuple[float, ...]) -> None:
```
Lasall, line 15: as type hint maybe we can just give Sequence[float]

PR #257, opened 2024-12-16
Docker usage

Lasall:
* Add EOS_CONFIG_DIR to set config dir (relative path to EOS_DIR or absolute path).
    - config_folder_path read-only
    - config_file_path read-only
 * Default values to support app start with empty config:
    - latitude/longitude (Berlin)
    - optimization_ev_available_charge_rates_percent (null, so model default value is used)
 * Fix some endpoints (empty data, remove unused params, fix types).
 * cacheutil: Use cache dir.
 * Support EOS_LOGGING_LEVEL environment variable to set log level.


src/akkudoktoreos/prediction/load_forecast.py
b0661, line 1: load_forecast.py can be replaced by LoadAkkudoktor prediction.  Maybe better to do that than to do the adaptions.

b0661:
The config handling in tests now look very much better. No interference anymore.

Very good that you removed the write to config within the simulation/ optimization.

b0661:
The comments on how to prevent the "Cannot compare tz-naive and tz-aware datetime-like objects" are just comments - no blocker from my side. I am working on some fixes that will also adjust this.

The comment on the change to the CacheFileStore is IMHO a blocker. I still think a writable /tmp volume shall be provided in the docker image. This is an assumption of Python. The application should not create work arounds because in case of a docker install the assumption is not fulfilled.

Lasall:
We could make /tmp rw, however I would argue that the cache should be stored in our cache dir. And then we wouldn't need rw on /tmp. I would argue it is not a workaround for docker but our system design where we store what data (and thanks to docker we noticed the problem).

Lasall:
(Same changes to visualize as in #281 to be able to mock the config object.)

src/akkudoktoreos/data/default.config.json
```diff
@@ -8,13 +8,13 @@
   "data_output_subpath": null,
   "elecprice_provider": null,
   "elecpriceimport_file_path": null,
-  "latitude": null,
```
NormannK, line 11: why did we set a default lat and long?

NormannK (approved):

src/akkudoktoreos/data/default.config.json
Lasall: So the system can be started and tested without any manual configuration.

NormannK (approved):

PR #262, opened 2024-12-17
translation of battery.py v3

NormannK:
translation of battery.py
this pr focuses on that single file.
test extended. old test included to proof function.


src/akkudoktoreos/devices/battery.py
```diff
@@ -10,345 +10,271 @@
...
         elif self.provider_id == "GenericBEV":
             self.prefix = "bev"
-        # Parameter initialisiation
+
         self.parameters = parameters
-        if hours is None:
-            self.hours = self.total_hours
-        else:
-            self.hours = hours
+        self.hours = hours if hours is not None else 24
```
Lasall, line 116: should probably get rid of the hours parameter in the init and use the config self.config. ... value

tests/test_batter.py
```diff
@@ -0,0 +1,106 @@
+import pytest
```
Lasall, line 1: filename test_battery.py

openapi.json
Lasall, line 1: duplicate

src/akkudoktoreos/devices/battery.py
Lasall: or maybe not, however we should replace the magic 24 by it

src/akkudoktoreos/devices/battery.py
```diff
@@ -10,345 +10,273 @@
...
+    max_charge_power_w: Optional[float] = max_charging_power_field()
 
 
-class EAutoParameters(BaseAkkuParameters):
-    entlade_effizienz: float = 1.0
-    start_soc_prozent: int = start_soc_prozent_field(
+class ElectricVehicleParameters(BaseBatteryParameters):
+    """Parameters specific to an electric vehicle (EV)."""
+
+    discharging_efficiency: float = 1.0
```
Lasall, line 68: this will probably also overwrite any description, so should use a field helper method with different default value

src/akkudoktoreos/devices/battery.py
```diff
@@ -10,345 +10,273 @@
 logger = get_logger(__name__)
 
 
-def max_ladeleistung_w_field(default: Optional[float] = None) -> Optional[float]:
+def max_charging_power_field(description: Optional[str] = "") -> float:
     return Field(
```
Lasall, line 14: we should always provide an API description, maybe if it's None here, use a default description later

src/akkudoktoreos/devices/battery.py
NormannK: lets address that in a new PR please

src/akkudoktoreos/devices/battery.py
NormannK: added

src/akkudoktoreos/devices/battery.py
```diff
@@ -10,345 +10,279 @@
 logger = get_logger(__name__)
 
 
-def max_ladeleistung_w_field(default: Optional[float] = None) -> Optional[float]:
+def max_charging_power_field(description: Optional[str] = None) -> Field:
```
Lasall, line 13: With more strict mypy this gives following error: ` Function "pydantic.fields.Field" is not valid as a type  [valid-type]` so should keep it the expected output type of the field

Lasall (commented):
Followup would be to cover the ev battery min/cur behavior with a test and handle the problem with pv battery (e.g. input validation)

src/akkudoktoreos/devices/battery.py
NormannK: ill set it to float. removing the optional since we have a default value

src/akkudoktoreos/devices/battery.py
```diff
@@ -10,345 +10,279 @@
...
-        default=default,
+        default=5000,
         gt=0,
-        description="An integer representing the charging power of the battery in watts.",
+        description=description,
     )
 
 
-def start_soc_prozent_field(description: str) -> int:
+def initial_soc_percentage_field(description: str) -> Field:
```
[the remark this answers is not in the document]
NormannK: ill set it to int.

Lasall (approved):

Issue #249, opened 2024-12-16
[BUG]: After pr #242 discharge is often 0 without result.Netzbezug_Wh_pro_Stunde >0


RobertHerter:
### Describe the issue:

Starting with pr#242 discharge_allowed[x] is often 0 without sense and
result.Netzbezug_Wh_pro_Stunde ist 0 at the same time.

### Version information:

pr#242

![image](https://github.com/user-attachments/assets/2aa65de6-f4db-4cf4-93a6-77de26d02797)

RobertHerter:
Request:
[...]

PR #268, opened 2024-12-19
Discharge hour Bug

drbacke:
Discharge hout bug, first fix https://github.com/Akkudoktor-EOS/EOS/issues/249 
- First self consumption rate due to possible high load peaks in inverter. TODO: self consumption must be calculated by load table 
- Self Consumption rate is FIXED!!! TODO!
- Penalty for not discharging removed


drbacke:
Notice for myself: 
- Faster Prediction (other interpolator) 
- Other routines (PV < consumption)

PR #271, opened 2024-12-21
Use Loadbalancer: Start api and html servers separately

Lasall:
* Fix loading config values from env arbitrary types (use json representation). Should be replaced by BaseSettings in the future (built-in support for env, nested config structure etc.).
    - Support derived ConfigEOS and tests for env loading.
 * Change fastapi server url to localhost:8503/api
    - fasthtml stays at localhost:8504/
 * fasthtml: Support development server (live reload)
 * Separate fasthtml/fastapi servers.
 * Makefile:
    - run     -> start both fastapi/fasthtml
    - run-dev -> start both fastapi/fasthtml (development mode, reload)
    - run-{api,html}     -> start either fastapi/fasthtml
    - run-{api,html}-dev -> start either fastapi/fasthtml (development mode, reload)
 * docker-compose:
    - use traefik as load balancer (requires read-access to docker.sock
      to read labels for auto-configure)
    - access to fasthtml at localhost:8503 and fastapi at localhost:8503/api
    - access treafik dashboard at localhost:8580

 * Note: fasthtml and fastapi start in separate containers (separation backend/frontend), so naturally fasthtml config display can only show the default config, except a volume mount is shared between the backend and frontend.

   In the future communication between the two should be handled via network requests (lose coupling) and they wouldn't necessarily share the same code base.


drbacke:
Could you Post this explanation into the Wiki. (For devs)

b0661:
What is the vision behind this setup? 

It looks quite complex for:
- an application server that is used to configure the rest server.
- an application and rest server that are run locally and only used by one user.

Lasall:
It is a standard backend/frontend architecture with divided startup for backend and frontend (for the dev setup no visible changes with make run{-dev,}, access at different adresses localhost:{8503,8504}). The docker setup ensures seemless access to both backend and frontend at the same adress (localhost:8503). I didn't add certificate support yet, could be easily done already with another reverse proxy on the server.

Now the vision for the frontend is GUI control over the backend functionality. Also we allow users to choose if they need the GUI app (should provide a headless compose file or use profiles in the current).

Currently we lose functionality in the frontend with the directory separation between backend/frontend (so the config app can only show default values). This could be fixed with a shared volume to the config dir. I can add that but in the future I would like to access all information from the backend API.

b0661:
Two containers make things quite complex. If you want to integrate that into Home Assistant you have to manage two containers, which is at least unusual for an Home Assistant integration.
Also load balancing can not be tested without docker containers. 
I don't see the benefit of load balancing for a single user in the Home Assistant example. Do you expect thousands of users for the servers?
The control of the application (fast_html) over the rest server (fast_api) can be achieved in the current setup also. So whats better/ easier in the load balancing setup?

Lasall:
I would disagree about the complexity (frontend/backend accessible over same url, with a few lines of code in compose.yml). Personally I use OpenHab where third party services have to be started separately and then can be added to the system by providing an URL to the service. When looking at the HA documentation about addons (and specifically evcc-addon/evcc), I can follow your thoughts about complexity of multiple services as the startup configuration is quite limited to make it most user-friendly (starting the third-party service from HA directly).

In my opinion it is important that EOS can be started standalone (for use with any system (HA, OpenHab, NodeRed, ..) which also allows users with limited hardware to split their services across devices. And it will make it more transparent to start each independent process separately and drop optional services (could be arguable done in an entrypoint.sh).
However I also agree we should support the very easy setup for HA by starting EOS from a single image. (Part of another PR when we provide HA-addon where we can make use of an entrypoint script to start both parts if wanted.)

This PR is less about load balancing, but about clear separation of backend/frontend. Also Treafik is used here just for convenience to allow users to access all EOS services at the same address.

b0661:
I am not against a two container solution. It may be appealing to deploy the application server to another machine than the REST server. I am against losing the solution for one container or non containerized installs.

My requirements would be:
1) The application server and the REST server shall be accessible by one IP address and port. The services/ business logic shall be distinguished by the path used.
2) It shall be possible to start the application server and the rest server independently
3) The install and deployment of the servers shall be possible with two containers or one container or non-containerized. Requirement 1) and 2) shall be fulfilled in all of the three deployment scenarios 

The current solution supports all 3 requirements:
1) The REST server proxies to the application server.
2) The REST server starts the application server only if the config for it is given. By setting this config to None the application server will not be started by the REST server. The application server can be started independently at any time. A special config flag for automatic startup of the application server by the REST server may be better here.
3)  Non containerized and one container solution already works. The two container solution may need the special config flag for automatic startup of the application server by the REST server. With the special config flag the IP address of the other server can still be provided but automatic startup can be prevented.

drbacke:
So the Basic question seems to be:
Shall we Develop an own GUI based in graphana ... or relay completely on HA.

A Basic lightweicht GUI would be very helpfull for developing and non HA Integrations.

b0661:
> Shall we Develop an own GUI based in graphana ... or relay completely on HA.

IMHO the question is more about deployment architecture and the split of concerns and the coupling between the REST server and the application server. 

The  fasthtml_server.py application is just one possible "basic lightweight GUI" application. You may at any time add a graphana, node red, Home Assistant, ... application.
   
> A Basic lightweicht GUI would be very helpfull for developing and non HA Integrations.

The lightweight GUI (aka. the fasthtml_server.py application) can even be directly integrated into Home Assistant (e.g. https://www.home-assistant.io/dashboards/iframe/). So it is very flexible to have this lightweight GUI.

drbacke:
> > Shall we Develop an own GUI based in graphana ... or relay completely on HA.
> 
> IMHO the question is more about deployment architecture and the split of concerns and the coupling between the REST server and the application server. 
> 
> The  fasthtml_server.py application is just one possible "basic lightweight GUI" application. You may at any time add a graphana, node red, Home Assistant, ... application.
>    
> > A Basic lightweicht GUI would be very helpfull for developing and non HA Integrations.
> 
> The lightweight GUI (aka. the fasthtml_server.py application) can even be directly integrated into Home Assistant (e.g. https://www.home-assistant.io/dashboards/iframe/). So it is very flexible to have this lightweight GUI.
> 
> 

Thanks for the explanation 😊

Issue #211, opened 2024-11-19
[BUG]: akku_soc_pro_stunde looks like 1 hour off 


RobertHerter:
### Describe the issue:

Showing the Results of an solution in Grafana to visualize teh solution, it look to me that akku_soc_pro_stunde ist one hour off.

In the Picture you could see, that on discharge 0 results the soc is one hour before.

![image](https://github.com/user-attachments/assets/00952a9d-b9fd-4d8b-8f52-416d535cffa3)

### Reproduceable code example:

```python
#Request
[...]

PR #276, opened 2024-12-22
Soc 1 hour shift fixed + some german -> english translations in ems

drbacke:


drbacke:
Fix https://github.com/Akkudoktor-EOS/EOS/issues/211

Issue #283, opened 2024-12-25
Discharge & Battery SoC = 0% bad optimization behaviour


drbacke:
If the battery still has residual energy at the beginning and is then drained relatively quickly, the subsequent discharge battery states no longer have any influence. The chance of charging the battery later is therefore extremely low. It would make sense to set the discharge_battery state at SoC = 0 constantly to 1 afterwards. This will probably solve the problem

PR #284, opened 2024-12-25
Awattar Price changed to more precise value; if SoC=0 discharge =1 forced. Exp Avg for siple Price Forecast

drbacke:
Awattar Price changed to more precise value; if SoC=0 discharge =1 forced. Exp Avg for siple Price Forecast


drbacke:
Fix https://github.com/Akkudoktor-EOS/EOS/issues/283

PR #302, opened 2024-12-30
Translated fastAPI endpoints from German to English

Dr4K4n:
Relates to #3 

I've started the journey to translate the code from German to English.
As this involves quite a lot of code I tried to make my commits as small as possible and plan to split the work into multiple pull requests. This first one is simply to get started. I've started with translating the API endpoints, which (in hindsight) might not have been the smartest, as it's quite a prominent place and will break any existing integrations 🥳. On the other hand, some time we will need to make this breaking change, so better now than later.

We could also keep the existing endpoint URLs (just as I did with /pvforecast, which I deprecated in favor of /pv_forecast). 

While getting to know the code I noticed some other small issues that I fixed on the way. Hope this doesn't mess up too much.

I also found another issue I want to further investigate (see reverted commit regarding pydantic @field_validator).

Looking forward to your feedback 😃

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ See [CONTRIBUTING.md](CONTRIBUTING.md).
 
 ## Installation
 
-The project requires Python 3.9 or newer. Currently there are no official packages or images published.
+The project requires Python 3.9 or newer. No official packages or images have been published yet.
 
 Following sections describe how to locally start the EOS server on `http://localhost:8503`.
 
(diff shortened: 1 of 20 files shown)


Lasall:
Thank you so much for digging in! Could you try to look at our current development branch feature/config-overhaul ? As far as I can see your changes mostly don't interfere with the changes there. And feel free to suggest/ask for any missing (development) documentation / improvements.

Lasall:
PR against current dev branch: https://github.com/Akkudoktor-EOS/EOS/pull/304

PR #338, opened 2025-01-04
documentation fixes

NormannK:
removes warnings from building


Lasall:
When looking at the files directly on github, their markdown does not support this syntax. I don't know if that's problematic or not as we enrich the markdown anyway.

PR #339, opened 2025-01-04
Initial readthedocs config

Lasall:
See example project at https://test-eos.readthedocs.io/en/latest/


docs/akkudoktoreos/serverapi.md
```diff
@@ -3,5 +3,5 @@
 # Server API
 
 ```{eval-sh}
-./scripts/generate_openapi_md.py | ./scripts/extract_markdown.py --input-stdin --start-line "**Version**:"
+python ${READTHEDOCS_REPOSITORY_PATH:-.}/scripts/generate_openapi_md.py | python ${READTHEDOCS_REPOSITORY_PATH:-.}/scripts/extract_markdown.py --input-stdin --start-line "**Version**:"
```
b0661, line 6: On a local build this will use the platform python and not the one from .venv. Did you try without python? Every script has a shebang and is executable. Doesn't this work on ReadTheDocs?

Lasall: Had to add python because readthedocs (also the execution path there is in the docs dir). I just realized when you pointed out the shebang, that I think it won't work on windows (i think the path there is .venv/Scripts/xyz).

Lasall: Ah, the path is probably why the shebang doesn't work on readthedocs (execution in .../docs so it doesn't have the .venv/xyz there). But I would probably change the shebang to just use python as there is a difference depending on platform.

b0661: Don't change the shebang, instead cd to  ${READTHEDOCS_REPOSITORY_PATH}/.. when on ReadTheDocs.

b0661: `python` won't work in my Python Windows installation at work. There I would have to use `py` or `px.exe`. Shebangs anyway do not work on Windows, except `#!/usr/bin/python`which is somehow handled by the Windows Python installation.

Lasall: I'll try how to make it work. I think the venv for readthedocs is also in a different dir.

b0661: Added #341 to avoid problems with ReadTheDocs (and Windows). You may use as a base for your change.

PR #349, opened 2025-01-06
feature: improve price forecast

NormannK:
Modifying elecpriceakkudoktor.py so it can hold up to 5 weeks of data.
Depending on the amount of data we will run different forecasts.
predictions_vs_true_ets_130
metrics_over_iterations_ets_cap_48
With 5 weeks we will use ets.


src/akkudoktoreos/prediction/elecpriceakkudoktor.py
```diff
@@ -184,80 +151,116 @@ def _update_data(self, force_update: Optional[bool] = False) -> None:
...
+        elif amount_datasets > 168:
+            prediction = self._predict_ets(
+                history, seasonal_periods=24, prediction_hours=7 * 24
+            )  # todo: add config values for prediction_hours
+        elif (
+            amount_datasets > 0
+        ):  # TODO might be a problem if amount_datasets is really low and we do the _cap_outliers.
+            prediction = self._predict_median(history, prediction_hours=7 * 24)
+        else:
+            assert False, "No data available"
```
b0661, line 207: Raise an exception, e.g. ValueError with more information. Also log the error to the log.

src/akkudoktoreos/prediction/elecpriceakkudoktor.py
```diff
@@ -184,80 +151,116 @@ def _update_data(self, force_update: Optional[bool] = False) -> None:
...
+        # Generate history array for prediction
+        history = np.array(
+            [
+                record.elecprice_marketprice_wh
+                for record in sorted(self.records, key=lambda r: r.date_time)
+                if record.elecprice_marketprice_wh is not None
+                and record.date_time
+                < highest_orig_datetime  # make sure we only real data for the prediction, so cant be newer then data from the api.
+            ]
+        )
```
b0661, line 189: Use `self.key_to_array(key="elecprice_marketprice_wh", end_datetime=highest_orig_datetime, fill_method="linear")` to get the array. This assures you get a value for every hour before highest_orig_datetime. None values will be interpolated.

src/akkudoktoreos/prediction/elecpriceakkudoktor.py
```diff
@@ -184,80 +151,116 @@ def _update_data(self, force_update: Optional[bool] = False) -> None:
...
+
+            existing_record = next((r for r in self.records if r.date_time == orig_datetime), None)
+            if existing_record:
+                # Update existing record
+                existing_record.elecprice_marketprice_wh = price_wh
+            else:
+                self.insert(
+                    0,
+                    ElecPriceDataRecord(date_time=orig_datetime, elecprice_marketprice_wh=price_wh),
+                )
```
b0661, line 178: Use `self.update_value(date=pred_datetime, key="elecprice_marketprice_wh", value=price)` to be sure to update existing data records (and do not create duplicates). If there is no existing data record it will be created automatically. No need to do that explicitly here.

src/akkudoktoreos/prediction/elecpriceakkudoktor.py
```diff
@@ -184,80 +151,116 @@ def _update_data(self, force_update: Optional[bool] = False) -> None:
...
+                self.insert(
+                    0,
+                    ElecPriceDataRecord(date_time=pred_datetime, elecprice_marketprice_wh=price),
+                )
+        history2 = np.array(  # TODO: remove return, only for debug, offset to see the difference
+            [
+                record.elecprice_marketprice_wh + 0.0002
+                for record in sorted(self.records, key=lambda r: r.date_time)
+                if record.elecprice_marketprice_wh is not None
+            ]
```
b0661, line 227: Use `self.key_to_array(key="elecprice_marketprice_wh", fill_method="linear") + 0.0002` to get history2.

src/akkudoktoreos/prediction/elecpriceakkudoktor.py
```diff
@@ -184,80 +151,116 @@ def _update_data(self, force_update: Optional[bool] = False) -> None:
...
+        predictions,
+        label="Predictions",
+        color="red",
+    )
+    plt.title("Predictions vs True Values for ets")
+    plt.xlabel("Time")
+    plt.ylabel("Price")
+    plt.legend()
+    plt.savefig("predictions_vs_true.png")
+    plt.close()
```
b0661, line 262: Move before main() in source code.

b0661 (changes requested):
Looks good. Some things can be done easier. Data records are always sorted by date unless you directly access self.records - so please never do.

src/akkudoktoreos/prediction/elecpriceakkudoktor.py
```diff
@@ -100,54 +86,6 @@ def _validate_data(cls, json_str: Union[bytes, Any]) -> AkkudoktorElecPrice:
...
-        else:
-            weights = self.elecprice_8days_weights_day_of_week[day_of_week]
-            prices_of_hour_masked: NDArray[Shape["24"]] = np.ma.MaskedArray(
-                prices_of_hour, mask=np.isnan(prices_of_hour)
-            )
-            price_weighted_mean = np.ma.average(prices_of_hour_masked, weights=weights)
-
-        return float(price_weighted_mean)
-
     @cache_in_file(with_ttl="1 hour")
```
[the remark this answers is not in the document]
NormannK: the api data is only updating once a day. problem is when we set the caching to long, we can ignore that update for some time.
worst case would be updatetime+cachetime.
1h would be on the save side.
id we would know when the api update is done we could check for the current time and cache 1day

tests/test_elecpriceakkudoktor.py
```diff
@@ -140,7 +120,7 @@ def test_update_data(mock_get, elecprice_provider, sample_akkudoktor_1_json, cac
 
     # Assert: Verify the result is as expected
     mock_get.assert_called_once()
-    assert len(elecprice_provider) == 49  # prediction hours + 1
```
NormannK, line 143: id why this is not 48+1 anymore. an advice is appreciated.

src/akkudoktoreos/prediction/elecpriceakkudoktor.py
```diff
@@ -184,80 +150,101 @@ def _update_data(self, force_update: Optional[bool] = False) -> None:
...
+                for i in range(len(prediction))
+            ],
+        )
+        self.key_from_series("elecprice_marketprice_wh", prediction_series)
+
+        # history2 = self.key_to_array(key="elecprice_marketprice_wh", fill_method="linear") + 0.0002
+        # return history, history2, prediction  # for debug main
+
+
+"""
```
b0661, line 215: Why do you make this a comment?

b0661 (changes requested):
Is good. Just the test must be updated to reflect the new logic.

tests/test_elecpriceakkudoktor.py
NormannK: nice catch. that was not intended. I'll calculate the needed_prediction_hours based on how many hours into the future we already have.

src/akkudoktoreos/prediction/elecpriceakkudoktor.py
NormannK: I'm unable to make it run bc of mypy if i keep it in.

PR #360, opened 2025-01-09
test_elecpriceakkudoktor bugfix

NormannK:
The tests for test_elecpriceakkudoktor.py fail for 2 cases when I run them all at once. However, when I restart the failing tests individually, the error disappears.

I traced the issue to the fact that the forecast hours being requested are negative. This means we somehow have more forecast data in the API response than we originally requested.

What should we do in such cases? Should we enforce the forecast to 0 hours? Should we remove the data from the record that goes too far into the future? How problematic is it if we have more data in the record than we expected, e.g., based on prediction_hours?

This fix would keep the data in the cache but does not try to run a prediction with negative or 0 needed_prediction_hours

### DIFF
--- a/src/akkudoktoreos/prediction/elecpriceakkudoktor.py
+++ b/src/akkudoktoreos/prediction/elecpriceakkudoktor.py
@@ -189,7 +189,7 @@ class ElecPriceAkkudoktor(ElecPriceProvider):
         )
 
         if needed_prediction_hours <= 0:
-            logger.error(
+            logger.warning(
                 f"No prediction needed. needed_prediction_hours={needed_prediction_hours}, prediction_hours={self.config.prediction_hours},highest_orig_datetime {highest_orig_datetime}, start_datetime {self.start_datetime}"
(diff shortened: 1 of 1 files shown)


src/akkudoktoreos/prediction/elecpriceakkudoktor.py
```diff
@@ -187,6 +187,13 @@ def _update_data(
             self.config.prediction_hours
             - ((highest_orig_datetime - self.start_datetime).total_seconds() // 3600)
         )
+
+        if needed_prediction_hours <= 0:
+            logger.error(
```
b0661, line 192: Is this an error? If not make it a debug message.

NormannK: i would say its an error. But its in error if the user stumbles on it, but its possible to force it e.g. with a test.
if the error is triggered its likely that start_datetime or prediction_hours is set wrong.

NormannK: changed to warning

Lasall (approved):

PR #361, opened 2025-01-09
Cleanup: parameters: extra=forbid, optimize: battery, inverter optional

Lasall:
* Don't allow extra fields for parameters (at least for now while changing API).
 * Allow both battery and inverter to be set optionally (atm optional battery not implemented, no API constraints).
 * inverter: Remove default max_power_wh
 * single_test_optimization: Add more cli-parameters
 * Docker: Fix entrypoint


src/akkudoktoreos/core/pydantic.py
```diff
@@ -35,6 +35,10 @@
 from akkudoktoreos.utils.datetimeutil import to_datetime, to_duration
 
 
+class ParametersBaseModel(BaseModel):
+    model_config = ConfigDict(extra="forbid")
+
+
```
b0661, line 41: If this only holds for optimization parameters, please move to optimization/optimizationabc.py and call it OptimizationParametersBase. Why do you use BaseModel? The custom PydanticBaseModel supports several custom types that are not supported by BaseModel.

NormannK (approved):

PR #362, opened 2025-01-10
Feature Branch: Generation Fitness plot

NormannK:
Added: A new plot displaying the minimum, maximum, and average fitness of the optimization process across generations.
Why: This plot helps analyze the performance of optimization runs:
- If the chart stabilizes towards the end with minimal changes, it indicates convergence.
- If the minimum fitness value reaches its optimal value quickly, the number of generations (ngen) can potentially be reduced to save computation time.
- Conversely, a lack of stabilization may suggest the optimization did not achieve satisfactory results.

graph for single_test_optimization.py:
image

### DIFF
--- a/single_test_optimization.py
+++ b/single_test_optimization.py
@@ -2,7 +2,6 @@
 
 import argparse
 import cProfile
-import logging
 import pstats
 import sys
 import time
@@ -18,8 +17,7 @@ from akkudoktoreos.optimization.genetic import (
 )
 from akkudoktoreos.prediction.prediction import get_prediction
 
-logger = get_logger(__name__)
-logger.setLevel(logging.DEBUG)  # set level for example report
+get_logger(__name__, logging_level="DEBUG")
 
 
 def prepare_optimization_real_parameters() -> OptimizationParameters:
--- a/single_test_optimization.py
+++ b/single_test_optimization.py
@@ -2,6 +2,7 @@
 
 import argparse
 import cProfile
+import logging
 import pstats
 import sys
 import time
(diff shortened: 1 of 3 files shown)


NormannK:
- plot is only generated if logging_level="DEBUG"
- added functions to add full pages with text (add_text_page)
- added function to print formatted json to a page (add_json_page)

known bug: using add_text_page and add_json_page will generated each a additional page with an empty chart. (idk why yet)

NormannK:
added a fixed seed if debug is enabled.
if debug or seed is set we produce the chart above with the seed number in the title

single_test_optimization.py
```diff
@@ -10,12 +11,16 @@
...
 from akkudoktoreos.config.config import get_config
 from akkudoktoreos.core.ems import get_ems
+from akkudoktoreos.core.logging import get_logger
 from akkudoktoreos.optimization.genetic import (
     OptimizationParameters,
     optimization_problem,
 )
 from akkudoktoreos.prediction.prediction import get_prediction
 
+logger = get_logger(__name__)
```
Lasall, line 21: could also use the logging_level parameter

src/akkudoktoreos/optimization/genetic.py
```diff
@@ -502,8 +512,10 @@ def optimize(
             for _ in range(10):
                 population.insert(0, creator.Individual(start_solution))
 
+        # logbook = tools.Logbook()
```
Lasall, line 515: unused code?

Lasall (commented):
Don't panic!

Lasall (approved):

single_test_optimization.py
NormannK: im sorry im lost here

src/akkudoktoreos/optimization/genetic.py
NormannK: removed

NormannK:
it would be great to find the bug with the empty pages before merging. or have a work around.

single_test_optimization.py
Lasall: get_logger(__name__, loglevel="DEBUG")

PR #364, opened 2025-01-12
Visualisation footer with current date and time

NormannK:
add a footer to each chart with the date and the version from pyproject.toml
image


src/akkudoktoreos/utils/visualize.py
```diff
@@ -1,6 +1,8 @@
+import datetime
```
b0661, line 1: To consistently use pendulum in Akkudoktor-EOS (and get rid of datetime) you could use:

```Python
from  akkudoktoreos.utils.datetimeutil import to_datetime
...
current_time = to_datetime(as_string=True)
```

If you want a special format of the string use the [Pendulum Formatter](https://pendulum.eustace.io/docs/#formatter), e.g.:

```Python
current_time = to_datetime(as_string="YYYY-MM-DD HH:mm:ss")
```

NormannK:
added a function to show time line data. prototype to replace the old one.
image
todo: change datetime to pendulum

NormannK:
[example_report.pdf](https://github.com/user-attachments/files/18400638/example_report.pdf)
[visualization_results.pdf](https://github.com/user-attachments/files/18400640/visualization_results.pdf)
I think its done.
Version number is static for now.

PR #371, opened 2025-01-13
Nested config, devices registry

Lasall:
* All config now nested.
    - Use default config from model field default values. If providers should be enabled by default, non-empty default config file could be provided again.
    - Environment variable support with EOS_ prefix and __ between levels, e.g. EOS_SERVER__EOS_SERVER_PORT=8503 where all values are case insensitive. For more information see: https://docs.pydantic.dev/latest/concepts/pydantic_settings/#parsing-environment-variable-values
    - Use devices as registry for configured devices. DeviceBase as base class with for now just initializion support (in the future expand to operations during optimization).
    - Strip down ConfigEOS to the only configuration instance. Reload from file or reset to defaults is possible.
 * Fix multi-initialization of derived SingletonMixin classes.

A few points still open:
- [ ] fix nested config in docs
- [ ] fix nested config in EOSdash
- [ ] remove now unnecessary prefix from config variables


b0661:
ConfigEOS already is a Singleton, so why do you want to make ir the only instance (which IT already is)?

Lasall:
You mean the extra check in `__init__`? The memory is the same but the `__init__` is still called every time a call to ConfigEOS() happens. (`__new__` returns (same) instance -> call to `__init__` again) Maybe there is a nicer way.

Issue #375, opened 2025-01-14
[BUG]: [feature] visualize Thread Error


drbacke:
### Describe the issue:

Crash of the server 
System: Ubuntu 20.04 / Proxmox

### Error message:

```shell
/home/andreas/eos/EOS/src/akkudoktoreos/utils/visualize.py:81: UserWarning: Starting a Matplotlib GUI outside of the main thread will likely fail.
  fig, axs = plt.subplots(rows, cols, figsize=(14, 7 * rows))
/home/andreas/eos/EOS/src/akkudoktoreos/utils/visualize.py:73: UserWarning: Starting a Matplotlib GUI outside of the main thread will likely fail.
  fig = plt.figure(figsize=(14, 10))
/home/andreas/eos/EOS/src/akkudoktoreos/utils/visualize.py:81: UserWarning: Starting a Matplotlib GUI outside of the main thread will likely fail.
  fig, axs = plt.subplots(rows, cols, figsize=(14, 7 * rows))
      INFO   192.168.1.135:47778 - "POST /optimize HTTP/1.1" 200
Exception ignored in: <function Image.__del__ at 0x71eecbf2e700>
Traceback (most recent call last):
  File "/home/andreas/anaconda3/lib/python3.11/tkinter/__init__.py", line 4080, in __del__
    self.tk.call('image', 'delete', self.name)
RuntimeError: main thread is not in main loop
      INFO   192.168.1.135:32824 - "GET /gesamtlast_simple?year_energy=8000 HTTP/1.1" 200
Exception ignored in: <function Variable.__del__ at 0x71eed006bb00>
Traceback (most recent call last):
  File "/home/andreas/anaconda3/lib/python3.11/tkinter/__init__.py", line 410, in __del__
    if self._tk.getboolean(self._tk.call("info", "exists", self._name)):
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: main thread is not in main loop
Exception ignored in: <function Variable.__del__ at 0x71eed006bb00>
Traceback (most recent call last):
  File "/home/andreas/anaconda3/lib/python3.11/tkinter/__init__.py", line 410, in __del__
    if self._tk.getboolean(self._tk.call("info", "exists", self._name)):
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: main thread is not in main loop
[...]

PR #377, opened 2025-01-14
fix for #375

NormannK:
we need to use the backend if we don't have a gui in matplotlib
https://matplotlib.org/stable/users/explain/figure/backends.html
we cant use python-tk ([tkinter](https://docs.python.org/3/library/tkinter.html#module-tkinter))


b0661:
Please add description to Code why Agg is necessary to prevent it deleted by accident again.

Lasall (approved):

PR #389, opened 2025-01-17
Nested config, devices registry 

Lasall:
* All config now nested.
    - Use default config from model field default values. If providers should be enabled by default, non-empty default config file could be provided again.
    - Environment variable support with EOS_ prefix and __ between levels, e.g. EOS_SERVER__EOS_SERVER_PORT=8503 where all values are case insensitive. For more information see: https://docs.pydantic.dev/latest/concepts/pydantic_settings/#parsing-environment-variable-values
    - Use devices as registry for configured devices. DeviceBase as base class with for now just initializion support (in the future expand to operations during optimization).
    - Strip down ConfigEOS to the only configuration instance. Reload from file or reset to defaults is possible.
 * Fix multi-initialization of derived SingletonMixin classes.

A few points still open:
- [x] fix nested config in docs
- [x] fix nested config in EOSdash
- [x] remove now unnecessary prefix from config variables
- [x] check for outdated documentation
- ~~move config x0, x1, etc. to lists~~ (do it in a future change: loadX_... into lists)

PR #400, opened 2025-01-21
Enhance secret configuration

### LINKED ISSUE #391: [ENH]: enhance privacy by limiting accuracy of lat and long values
### Link to discussion and related issues

Any latitude and longitude values managed by the system should be restricted to an accuracy level that prevents exposing the user's precise location when transmitted via an API or included in outputs, such as debug data in a GitHub issue. If these values are stored in a configuration file, the stored data should reflect the same limited accuracy.

Accuracy versus decimal places at the equator
```python
decimal  degrees    distance
places
-------------------------------  
0        1.0        111 km
1        0.1        11.1 km
2        0.01       1.11 km
3        0.001      111 m
4        0.0001     11.1 m
5        0.00001    1.11 m
6        0.000001   0.111 m
7        0.0000001  1.11 cm
8        0.00000001 1.11 mm
```
I propose a accuracy of 1.11 km so two decimal places.

### Proposed implementation

```python

```

b0661:
Enhance configuration to support secret configuration values. Secret configuration values are held in a secrets file.

In the "normal" configuration the secret configuration can be referenced by the value "!secret <secret_key>".

The secret can only be accessed by config.secret("\<key\>") which follows the indirection if there is any, otherwise it returns the current value. If the secret value is accessed by  config.\<key\> and there is a secret redirection then "!secret <secret_key>" will be returned, thus hiding the secret.

Closes #391

Should wait for #389 to cover nested configuration.


b0661:
We switch to nested config.

PR #418, opened 2025-01-26
Data prefetch for ems

NormannK:
Added data and object pre-fetching for `def simulate`.
Some code cleanup
10% better runtime for single_test_optimization.py.

extended tests and some comment maintenance there.

### DIFF
--- a/src/akkudoktoreos/core/ems.py
+++ b/src/akkudoktoreos/core/ems.py
@@ -411,9 +411,9 @@ class EnergieManagementSystem(SingletonMixin, ConfigMixin, PredictionMixin, Pyda
             costs_per_hour[hour_idx] = energy_consumption_grid_actual * hourly_electricity_price
             revenue_per_hour[hour_idx] = energy_feedin_grid_actual * hourly_energy_revenue
 
-        total_cost = np.sum(costs_per_hour)
-        total_losses = np.sum(losses_wh_per_hour)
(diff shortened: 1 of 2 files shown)


src/akkudoktoreos/core/ems.py
```diff
@@ -276,53 +276,50 @@ def simulate_start_now(self) -> dict[str, Any]:
...
+        pv_prediction_wh = np.array(self.pv_prediction_wh)
+        elect_price_hourly = np.array(self.elect_price_hourly)
+        ev_charge_hours = np.array(self.ev_charge_hours)
+        ac_charge_hours = np.array(self.ac_charge_hours)
+        dc_charge_hours = np.array(self.dc_charge_hours)
+        elect_revenue_per_hour_arr = np.array(self.elect_revenue_per_hour_arr)
+
+        # Fetch objects
+        battery = self.battery
+        assert battery  # to please mypy
```
Lasall, line 313: I'm not quite sure if the system already supports no battery, but we should think in that direction.

NormannK: it does not. i did remove several checks for the battery after that.
i can not think of a situation for eos without a battery. and if we really would run without i would make a fake battery with 0 capacity. 
the same is true for ev currently. we should make it usable without ev. i did keep all the checks for ev though.

Lasall (approved):

PR #423, opened 2025-01-26
REST: Allow setting single config value

Lasall:
* /v1/config/{path} supports setting single config value (post body). Lists are supported as well by using the index:
    - general/latitude (value: 55.55)
    - optimize/ev_available_charge_rates_percent/0 (value: 42)

   Whole tree can be overriden as well (no merge):
    - optimize/ev_available_charge_rates_percent (value: [42, 43, 44])

### DIFF
--- a/src/akkudoktoreos/core/pydantic.py
+++ b/src/akkudoktoreos/core/pydantic.py
@@ -51,6 +51,70 @@ def merge_models(source: BaseModel, update_dict: dict[str, Any]) -> dict[str, An
     return merged_dict
 
 
+def access_nested_value(
+    model: BaseModel, path: str, setter: bool, value: Optional[Any] = None
+) -> Any:
+    """Get or set a nested model value based on the provided path.
+
+    Supports string paths (with '/' separators) or sequence paths (list/tuple).
+    Trims leading and trailing '/' from string paths.
+
+    Args:
+        model (BaseModel): The model object for partial assignment.
+        path (str): The path to the model key (e.g., "key1/key2/key3" or key1/key2/0).
+        setter (bool): True to set value at path, False to return value at path.
+        value (Optional[Any]): The value to set.
+
+    Returns:
+        Any: The retrieved value if acting as a getter, or None if setting a value.
+    """
+    path_elements = path.strip("/").split("/")
+
+    cfg: Any = model
+    parent: BaseModel = model
+    model_key: str = ""
+
+    for i, key in enumerate(path_elements):
+        is_final_key = i == len(path_elements) - 1
+
+        if isinstance(cfg, list):
+            try:
+                idx = int(key)
+                if is_final_key:
+                    if not setter:  # Getter
+                        return cfg[idx]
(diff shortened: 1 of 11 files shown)


src/akkudoktoreos/server/eos.py
```diff
@@ -302,6 +304,55 @@ def fastapi_config_put(settings: SettingsEOS) -> ConfigEOS:
...
+            cfg[idx] = value
+        except (ValueError, IndexError):
+            raise HTTPException(status_code=400, detail=f"Invalid list index at {path}: {key}")
+    elif hasattr(cfg, key):
+        try:
+            setattr(cfg, key, value)
+        except Exception as e:
+            raise HTTPException(status_code=400, detail=f"Error on update of configuration: {e}")
+    else:
+        raise HTTPException(status_code=404, detail=f"Key '{key}' not found in configuration.")
```
b0661, line 351: Could that become a method of ConfigEOS? 

E.g.
```python
def set_config_value(self, path: Union[Sequence, str], value) -> None:

def get_config_value(self, path: Union[Sequence, str]) -> Any:
```

src/akkudoktoreos/core/pydantic.py
```diff
@@ -51,6 +52,71 @@ def deep_update(source_dict: dict[str, Any], update_dict: dict[str, Any]) -> dic
     return merged_dict
 
 
+def access_config_value(
+    config: BaseModel, path: Union[str, Sequence[str]], setter: bool, value: Optional[Any] = None
+) -> Any:
+    """Get or set a configuration value based on the provided path.
+
+    Supports string paths (with '/' separators) or sequence paths (list/tuple).
+    Trims leading and trailing '/' from string paths.
```
b0661, line 61: Is this specific to config? If not, you may rename to something more generic.

E.g.:
```python
def access_nested_value(
    model: BaseModel, path: Union[str, Sequence[str]], setter: bool, value: Optional[Any] = None
)
```

src/akkudoktoreos/core/pydantic.py
```diff
@@ -51,6 +52,71 @@ def deep_update(source_dict: dict[str, Any], update_dict: dict[str, Any]) -> dic
...
+    model: BaseModel, path: Union[str, Sequence[str]], setter: bool, value: Optional[Any] = None
+) -> Any:
+    """Get or set a nested model value based on the provided path.
+
+    Supports string paths (with '/' separators) or sequence paths (list/tuple).
+    Trims leading and trailing '/' from string paths.
+
+    Args:
+        model (BaseModel): The model object for partial assignment.
+        path (Union[str, Sequence[str]]): The path to the model key (e.g., "key1/key2/key3" or ["key1", "key2", "key3"]).
```
b0661, line 65: What about ints in the Sequence to denote list elements?

Lasall: Good point, however it's not used atm and I didn't write any tests for it, so I would remove the Sequence support.

PR #426, opened 2025-01-26
Config: Don't fail on config error, fix Windows config dir

Lasall:
* Support setting logger level (env) before config load.


Lasall:
still crashes if env/dotenv has an error.

NormannK:
how am i supposed to test this?

Issue #387, opened 2025-01-17
[BUG]: Time-Axis in visualization_results.pdf are wrong (an now Marker)


RobertHerter:
### Describe the issue:

With new master the visualization_results.pdf Time-Axis is wrong an starts at 00:00 for "Now".

![Image](https://github.com/user-attachments/assets/0c23df42-94f2-4bfe-b943-06c560273cf0)

### Reproduceable code example:

```python

```

### Error message:

```shell

```

### Version information:

Master

NormannK:
I need to make it clearer how to interpret the graph. The red dashed line represents the current time ("now") and corresponds to the server time, adjusted for the configured time zone. The x-axis at the top shows the number of hours since the first recorded value, as it did in the old graphs. The x-axis at the bottom represents the actual time.

In the future we will also show past data before now, that is why the indicator is helpful.

Does that help?

RobertHerter:
Hello Normann,

i agree that the indictor of now is helpfull, but the lower x starts at 8:00 (witch is the time i run the solution) but the Graph and top x shows starting from 00:00. Looking at the bottom x-axis the interpretation is that i have maximum pv around midnight ;-).

RobertHerter:
Last run at 18:15 CET

[visualization_results.pdf](https://github.com/user-attachments/files/18457866/visualization_results.pdf)

NormannK:
from the request right the first item is this:

```
"pv_prognose_wh": [
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                146.20268649912595,
                512.847910990359,
                856.2969903403294,
                1070.1335984765183,
                816.3972436141687,
                591.9229498875227,
                465.4621868374305,
                196.70586731633315,
                26.431601158551082,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                366.8,
                1820.15,
                3341.025,
                4550,
                5138.625,
                4983.825,
                4015.475,
                2760.625,
                1240.2,
                15.375,
                0,
                0,
                0,
                0,
                0,
                0,
            ],
```
which does fit to the PV Forecast chart. How did you generate the PV Forecast data?

NormannK:
From your request the first item is this:

```
"pv_prognose_wh": [
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                0,
                146.20268649912595,
                512.847910990359,
[...]

PR #430, opened 2025-01-28
visualize: fix timestamps on diagramms

celle1234:
I tried to fix the problem with the timestamps in the diagrams. The discussion you can find in #387.

### DIFF
--- a/src/akkudoktoreos/utils/visualize.py
+++ b/src/akkudoktoreos/utils/visualize.py
@@ -13,6 +13,7 @@ import pendulum
 from matplotlib.backends.backend_pdf import PdfPages
 
 from akkudoktoreos.core.coreabc import ConfigMixin
+from akkudoktoreos.core.ems import EnergieManagementSystem
 from akkudoktoreos.core.logging import get_logger
 from akkudoktoreos.optimization.genetic import OptimizationParameters
 from akkudoktoreos.utils.datetimeutil import to_datetime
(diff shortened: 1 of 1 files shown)


NormannK:
[visualization_results.pdf](https://github.com/user-attachments/files/18575604/visualization_results.pdf)
from single_test_optimization.py for discussion

Lasall:
Thank you for your PR! Currently the optimization endpoint (and therefore the visualization) expect the forecast data starting at 00:00 of the current day + x prediction hours (currently 48). The optimization starts at start_hour (default rounded down current hour, e.g. 16:42 -> 16) and goes until optimization hours (currently 48) which starts at 00:00 of the current day just like the predictions (e.g. start_hour = 16, optimization_hours = 48 - > optimization will go until end of the next day).

celle1234:
@Lasall Yes I understand this in the same way. 

For the PR it is important, that the timezone in EOS.config.json is set. Otherwise the axis timeline does not match.

Lasall:
[new_optimize_result_2_full.pdf](https://github.com/user-attachments/files/18609327/new_optimize_result_2_full.pdf)
from test artifact.

Maybe you could also adjust the relative times "hours since start" to be the same between predictions and optimization?Currently it could be confusing the prediction goes 0-48 and optimization 0-38 (even though optimization_hours is set to 48).

celle1234:
Did you mean something like this?
[visualization_results.pdf](https://github.com/user-attachments/files/18623817/visualization_results.pdf)

NormannK:
> Did you mean something like this? [visualization_results.pdf](https://github.com/user-attachments/files/18623817/visualization_results.pdf)

after some code digging I think this is "more" correct. https://github.com/Akkudoktor-EOS/EOS/blob/1a2da7636b7149169b44d37dd5103409a2a2db5e/src/akkudoktoreos/core/ems.py#L128
This line does set the start time to the start of the current hour (if not defined). But imo it should be the start of the next hour, but that's something for an other discussion.

celle1234:
I added the changes.

Lasall:
Looks good to me, probably need some fine tuning with the optimization diagram (offset +1) but in a different PR (I think that's also what you meant Normann).

Lasall (approved):

NormannK (approved):

Lasall:
Thank you!

PR #453, opened 2025-02-11
Initial "optimize" documentation

XHalamaX:
Here is the first version of the optimize API. I also added the "raw" data for the graphic so it can be easliy imported to https://excalidraw.com/
hope this is ok


b0661:
The Markdown format has to be adapted a little bit to follow common Markdown conventions. See #454 for an example.

This pull request is already part of the feature branch feature/documentation. The usual way is to finish the feature and then merge the feature branch into main. Is there a special reason to merge this pull request before the feature branch?

XHalamaX:
@b0661

I will add more and fix some issues, but I don't know when I'll have time for this.  

From my perspective, we should publish documentation updates sooner rather than later so users don’t waste time or get frustrated. But that's just my point of view.

PR #458, opened 2025-02-15
EOSdash: Shutdown on EOS termination.

Lasall:
* Use Popen.terminate to stop process started in lifespan.


b0661:
Kills EOSdash on shutdown and restart endpoint. Doing server admin from EOSdash would be made impossible. At least when triggering these endpoints a parameter should allow to prevent killing of EOSdash.

PR #461, opened 2025-02-16
pre-commit update and ignore changes

NormannK:
updates components of pre-commit.
changes to pyproject.toml to ignore  "error: Unused "type: ignore" comment  [unused-ignore]"
type: ignore used for linux and windows commands

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,7 +84,7 @@ testpaths = [ "tests", ]
 files = ["src", "tests"]
 exclude = "class_soc_calc\\.py$"
 check_untyped_defs = true
-warn_unused_ignores = false
+warn_unused_ignores = true
 
 [[tool.mypy.overrides]]
 module = "akkudoktoreos.*"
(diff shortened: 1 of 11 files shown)


pyproject.toml
```diff
@@ -84,7 +84,7 @@ testpaths = [ "tests", ]
 files = ["src", "tests"]
 exclude = "class_soc_calc\\.py$"
 check_untyped_defs = true
-warn_unused_ignores = true
+warn_unused_ignores = false
```
Lasall, line 87: According to https://github.com/python/mypy/issues/8823 it is possible to add `ignore[unused-ignore]`, e.g. `# type: ignore[attr-defined,unused-ignore]`. This seem to work well on Linux when I add it to eos.py and test_server.py (both Linux and Windows code paths).

tests/conftest.py
```diff
@@ -311,7 +311,7 @@ def cleanup_eos_eosdash():
             # Windows does not provide SIGKILL
             sigkill = signal.SIGTERM
         else:
-            sigkill = signal.SIGKILL
+            sigkill = signal.SIGKILL  # type: ignore
```
NormannK, line 314: this does only work with a general ignore. please confirm its correct.

PR #490, opened 2025-03-23
Update dependabot.yml to also include the feature branch

NormannK:
sadly wildcard paths are not possible

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,7 +5,16 @@
 
 version: 2
 updates:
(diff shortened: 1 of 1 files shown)


b0661:
Hi @NormannK, @Lasall will prepare for the merge of feature/config-nested to main. I think this obsoletes this PR. Would be nice though if we could have a wildcard like feature/*.

Please close if you agree.

NormannK:
sadly wildcard paths are not possible

PR #523, opened 2025-04-21
fix: mitigate ReDoS in to_duration via max input length check (closes #494)

### LINKED ISSUE #494: [BUG]: ReDOS Vulnerability based on User Input
### Summary of the Vulnerability:

The regular expression `(\d+)\s*(days?|hours?|minutes?|seconds?)` is vulnerable to ReDoS because it allows the possibility of catastrophic backtracking when certain inputs are given. Specifically, if a user provides a highly repetitive string such as `'0'.repeat(54773) + 'h0day'`, it could cause excessive backtracking, leading to performance degradation or denial of service. This type of attack can overload the server and cause it to hang, potentially crashing the system.

### Example of Exploit:

The exploit works by submitting the following input for the `interval` parameter:

```python
'0'.repeat(54773) + 'h0day'
```

This string triggers a catastrophic backtracking scenario due to the regular expression, causing the system to expend excessive resources while trying to match the pattern.

### Impact:

- **Performance Degradation:** The server will be forced into a heavy computation cycle, leading to a significant slowdown.
- **Denial of Service (DoS):** In extreme cases, it may cause the application to become unresponsive, leading to downtime for the service.

### Fix:

To mitigate the ReDoS vulnerability, the regular expression should be optimized to avoid backtracking. Here are a few recommendations:

1. **Limit Input Length:** Restrict the maximum allowed length of the `interval` parameter to avoid overly long strings.
2. **Use Non-Backtracking Regular Expressions:** Use a more efficient pattern or regex engine that avoids catastrophic backtracking, such as using a fixed-length format for time units.
3. **Pre-process User Input:** Before passing the input to the regex, ensure that it follows a predefined format, such as allowing only numerical input or specific predefined intervals.

Fixed Regex : ^\d+\s*(days?|hours?|minutes?|seconds?)$

```python
import re

def to_duration(interval: str) -> int:
    # Allow only specific units and prevent catastrophic backtracking.
    if not re.match(r'^\d+\s*(days?|hours?|minutes?|seconds?)$', interval):
        raise ValueError(f"Invalid interval format: {interval}")
    # Continue processing the valid interval...
```

### Conclusion:

By tightening the regular expression and enforcing stricter input validation, you can prevent ReDoS vulnerabilities in your FastAPI application. Always ensure that user input is properly sanitized and validated before being processed.

```python

```

### Error message:

```shell

```

### Version information:

latest

ChristianHohlfeld:
This PR fixes a potential ReDoS vulnerability in the `to_duration()` function.

Instead of restructuring the regex (which could break multi-part duration strings like "2 days 5 hours"), this fix introduces a clear and robust mitigation: a maximum input length check at the start of the function.

Includes:
- Max length check (`MAX_DURATION_STRING_LENGTH = 350`)
- Dedicated unit test to validate correct error and boundary behavior

Thanks @aydinnyunus for reporting and analyzing the issue (#494).

### DIFF
--- a/src/akkudoktoreos/utils/datetimeutil.py
+++ b/src/akkudoktoreos/utils/datetimeutil.py
@@ -34,6 +34,7 @@ from timezonefinder import TimezoneFinder
 from akkudoktoreos.core.logging import get_logger
 
 logger = get_logger(__name__)
+MAX_DURATION_STRING_LENGTH = 350
 
 
 @overload
@@ -287,6 +288,11 @@ def to_duration(
             "second": 1,
         }
 
+        # Mitigate ReDoS vulnerability (#494) by checking input string length.
+        if len(input_value) > MAX_DURATION_STRING_LENGTH:
+            raise ValueError(
+                f"Input string exceeds maximum allowed length ({MAX_DURATION_STRING_LENGTH})."
+            )
         # Regular expression to match time components like '2 days', '5 hours', etc.
         matches = re.findall(r"(\d+)\s*(days?|hours?|minutes?|seconds?)", input_value)
 
--- a/tests/test_datetimeutil.py
+++ b/tests/test_datetimeutil.py
@@ -1,10 +1,13 @@
 """Test Module for pendulum.datetimeutil Module."""
 
+import re
+
 import pendulum
 import pytest
 from pendulum.tz.timezone import Timezone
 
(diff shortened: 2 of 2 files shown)


Lasall:
Thank you for the provided hardening fix plus the unit test. Please run the pre-commit hook to fix the formatting and also translate the comments to English.

PR #546, opened 2025-05-10
Activation of venv is needed

SchaafAlexander:
On my windows device the following code was additionally needed to start the virtial environment:
.venv\Scripts\Activate

So I added it to the installation documentation.

Maybe the same is needed on Linux, but not tested from my side.


NormannK:
very nice. I was the only one using windows :-) and i always wondered why I had to do that manually.

Issue #381, opened 2025-01-14
[ENH]: Add new energyprice provider energy-charts.info


NormannK:
### Link to discussion and related issues

```python
replace akkudoctor energyprice api by energy-charts.info
https://api.energy-charts.info/#/prices/day_ahead_price_price_get
```

### Proposed implementation

```python
basically copy ElecPriceAkkudoktor to ElecPriceenergycharts
```

NormannK:
can be done in combination #365 
if tibber is used tax values should be set/checked by tibber data and energy-charts.info can be used as fall back information.
maybe even if there is more up to date data available there.

b0661:
Do Not replace, but add a new prediction. Sich you can Switch predictions by config.prediction_provider.

b0661:
Reopen due to stale policy change for enhancements.

b0661:
Now in main.

PR #590, opened 2025-06-16
Added new electricity price provider: Energy-Charts #381

redmoon2711:
Added new electricity price provider: Energy-Charts

- Implemented Energy-Charts as a new data source for electricity prices
- Based on existing elecpriceakkudoktor.py implementation
- Calculation changed to (maketprice + charges) * VAT  for better results
- All tests passed successfully

Have fun with it :-)

### DIFF
--- a/src/akkudoktoreos/prediction/elecpriceenergycharts.py
+++ b/src/akkudoktoreos/prediction/elecpriceenergycharts.py
@@ -11,16 +11,14 @@ from typing import Any, List, Optional, Union
 import numpy as np
 import pandas as pd
 import requests
+from loguru import logger
 from pydantic import ValidationError
 from statsmodels.tsa.holtwinters import ExponentialSmoothing
 
 from akkudoktoreos.core.cache import cache_in_file
-from akkudoktoreos.core.logging import get_logger
 from akkudoktoreos.core.pydantic import PydanticBaseModel
 from akkudoktoreos.prediction.elecpriceabc import ElecPriceProvider
 from akkudoktoreos.utils.datetimeutil import to_datetime, to_duration
-
-logger = get_logger(__name__)
 
 
 class EnergyChartsElecPrice(PydanticBaseModel):
@@ -90,12 +88,14 @@ class ElecPriceEnergyCharts(ElecPriceProvider):
             - add the file cache again.
         """
         source = "https://api.energy-charts.info"
-        assert self.start_datetime  # mypy fix
+        if not self.start_datetime:
+            raise ValueError(f"Start DateTime not set: {self.start_datetime}")
+
         # Try to take data from 5 weeks back for prediction
         date = to_datetime(self.start_datetime - to_duration("35 days"), as_string="YYYY-MM-DD")
         last_date = to_datetime(self.end_datetime, as_string="YYYY-MM-DD")
         url = f"{source}/price?bzn=DE-LU&start={date}&end={last_date}"
-        response = requests.get(url)
+        response = requests.get(url, timeout=10)
         logger.debug(f"Response from {url}: {response}")
         response.raise_for_status()  # Raise an error for bad responses
         energy_charts_data = self._validate_data(response.content)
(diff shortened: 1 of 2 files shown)


src/akkudoktoreos/prediction/elecpriceenergycharts.py
```diff
@@ -0,0 +1,211 @@
...
+            error_msg = f"Highest original datetime not available: {highest_orig_datetime}"
+            logger.error(error_msg)
+            raise ValueError(error_msg)
+
+        # some of our data is already in the future, so we need to predict less. If we got less data we increase the prediction hours
+        needed_hours = int(
+            self.config.prediction.hours
+            - ((highest_orig_datetime - self.start_datetime).total_seconds() // 3600)
+        )
+        print(f"EOS CONFIG {self.config.prediction.hours}")
```
b0661, line 185: Please use logger instead of print.

src/akkudoktoreos/prediction/elecpriceenergycharts.py
```diff
@@ -0,0 +1,211 @@
...
+        date range. The response data is parsed and returned as JSON for further processing.
+
+        Returns:
+            dict: The parsed JSON response from Energy-Charts API containing forecast data.
+
+        Raises:
+            ValueError: If the API response does not include expected `electricity price` data.
+
+        Todo:
+            - add the file cache again.
```
b0661, line 88: What does that mean - add file cache again?

b0661 (changes requested):
Looks really good. Please add a test.

src/akkudoktoreos/prediction/elecpriceenergycharts.py
redmoon2711: I'm also not sure what exactly "add file cache again" is supposed to mean. I took those lines from elecpriceakkudoktor.py.
But I could imagine that the idea is to check which data is already in the cache before making the request, and then either skip the request or perform it with a shorter start date.

For example:
- If data is available up to the day after tomorrow at 00:00, then skip the request.
- If data is available up to tomorrow at 00:00 and it's before 1:00 PM, then skip the request.
- If data is available up to tomorrow at 00:00 and it's after 1:00 PM, then fetch new data for tomorrow.
- If no data is available, then fetch all data.

b0661: Just remove what you do not know of. Otherwise make clear what you may add in the future.

b0661:
I have seen that you added the test. Thank you, that helps a lot.

 One last thing, could you also add documentation to predictions.md, so others can get information about this provider.

b0661:
I assume the failing ci tests are unrelated to your PR. We access in test to a real internet server, which is currently down. That is the reason why this oddly test bug now emerged.

.idea/EOS_main.iml
b0661, line 1: Do not check in workspace files from your IDE (I assume Pycharm). Remove the the whole .idea directory from the pull request. You may add ".idea" to .gitignore to avoid this in the future.

b0661:
> I assume the failing ci tests are unrelated to your PR. We access in test to a real internet server, which is currently down. That is the reason why this oddly test bug now emerged

The server is up again.

Please have a look at test_prediction.py. Your change is triggering test faults as you added the new provider without adapting the test.

You may run the tests before you push your changes. "make test" or ".venv/bin/pytest -vs --cov src --cov-report term-missing" is your friend.

.idea/EOS_main.iml
redmoon2711: oh... ok i remove the .idea directory

src/akkudoktoreos/prediction/elecpriceenergycharts.py
NormannK: imo this is blocking the merge.
very good job so far

NormannK (approved):

b0661 (approved):

PR #607, opened 2025-06-25
Optimization of the Inverter class to speed up scr calculation

redmoon2711:
Hi, 

in the Inverter class, the scr (self_consumption_predictor) is recalculated for every hour and every generation. However, since the value is often identical, the runtime can be significantly reduced by caching the result in a lookup table.
In my case, this optimization nearly halves the execution time.

Try it out and enjoy the result :-)


NormannK:
good point.
I'm only worried that for every unique (consumption, generation) pair you create a new entry. Especially since those are floats that could lead to a very high number of entries. 
Do you think it would be worth it to round the values to like 5 digits and/or use a TTL for the entries? Like log the last time used and discard everything with a certain age?

redmoon2711:
I've been using this optimization for several months now and haven't observed any negative impact.
The Inverter class is re-initialized with every EOS run anyway.

In my tests, I observed that the _calculate_scr function is called over 600,000 times.
However, the lookup table itself contains only a small number of entries, since solar generation and consumption vary only slightly. Rounding the values to 2 decimal places had little to no effect.

PR #727, opened 2025-10-30
chore: distinguish prediction data from solution data in generic optimization solution

b0661:
Distinguish prediction data from solution data in generic optimization solution.
Add some predictions that may be of interest to understand the solution.


b0661:
Will be replaced by fix for zero load values.

Issue #798, opened 2025-12-05
[BUG]: Laden/Entladen der Batterie nicht immer korrekt


FidoWue:
### Describe the issue:

Herzlichen Dank für EOS. 

Ich setze EOS seit ca. 8 Wochen produktiv mit einer 21,2 kW peak PV-Anlage, 23 kWh Batterien und einem dynamischen Stromtarif ein. EOS v0.2.0 (aktuell in docker) ist über EOS-connect 0.2.29.198-develop in HomeAssistant integriert.

Die Vorhersage für  das Laden und Entladen der Batterie ist immer wieder etwas merkwürdig. Ich weiß nicht, ob dies dem frühen Entwicklungsstand geschuldet ist (in dem Fall schließt einfach das Ticket) oder ob bei mir etwas schief läuft.

![Image](https://github.com/user-attachments/assets/490233cd-e3d2-4a29-adbc-254fdfacb3b9)

Auf dem Screenshot sind für mich mehrere Probleme zu sehen:
- Das Entladen der Batterie bis 23:00 Uhr wird mehrmals unterbrochen
- 12% ist die untere Grenze für die Entladung. Ab 19:00 Uhr soll die Batterie weiter entladen werden.
- Um 11:00 Uhr wird die Batterie mit teuerem Strom, geladen, anstatt um 07:00 weiter zu Laden.
- 80% ist als Maximium für die Ladung konfiguriert. Ab 14:00 Uhr soll die volle Batterie weiter geladen werden.

Anbei die Configs:
[EOS.config.json](https://github.com/user-attachments/files/23965775/EOS.config.json)
[EOS-Connect.txt](https://github.com/user-attachments/files/23965783/EOS-Connect.txt)

### Reproduceable code example:

```python

```

### Error message:

```shell

```

### Version information:

EOS v0.2.0 (aktuell in docker)

b0661:
> I can see several problems in the screenshot:
>
>- The battery discharge until 11:00 p.m. is interrupted several times.
>- 12% is the lower limit for discharge. From 7:00 p.m. onwards, the battery should continue to discharge.
>- At 11:00 a.m., the battery is charged with expensive electricity instead of continuing to charge at 7:00 a.m.
>- 80% is configured as the maximum charge. From 2:00 p.m. onwards, the full battery should continue to charge.

The genetic algorithm tries to find a good solution, this is not always the best one (but should almost be the best one in most cases). Good is defined as the lowest total costs.

[...]

Issue #844, opened 2026-01-09
[Q] I am not able to get any useful optimization out of EOS yet.


b0661:
@b0661 I can't provide that, as I am not able to get any useful optimization out of EOS yet. Currently, it is a lottery game, which input I should pick next to provide to EOS to possibly get a useful optimization out of it.
I tried providing household load to get a prediction of the household load, which was a failure. Remember my reported issue with the by factor 1000 changing values after providing household load? This is still not solved.

_Originally posted by @kleini in https://github.com/Akkudoktor-EOS/EOS/issues/829#issuecomment-3729100821_

b0661:
> I tried providing household load to get a prediction of the household load, which was a failure. Remember my reported issue with the by factor 1000 changing values after providing household load? This is still not solved.

Please see PR #826 which was recently merged.

b0661:
Hello @kleini, the easiest way to initialize EOS after installation is to activate automatic optimization (ems.mode = OPTIMIZATION) with a test interval of, for example, 120 seconds (ems.interval=120). During the first optimization run, demo values ​​will be entered into the configuration, and after a while, you should be able to retrieve the predicted values ​​with EOSdash. Afterward, you can adjust the configuration to your needs. Don't forget to save!

There's also a thread about this in the forum: https://akkudoktor.net/t/installation-eos/35390/3

You may provide the type of system you are planning to use (Home Assistant, NodeRed, stand alone, ...) to provide you with more targeted answers.

kleini:
Is it somewhere written down in the documentation, that the best starting point is to just set these two configuration options and than start to get an optimization plan?

b0661:
> Is it somewhere written down in the documentation, that the best starting point is to just set these two configuration options and than start to get an optimization plan?

No, because this was introduced in the development version and it is unclear whether this will become the default.

You may try and see if it shall become the default (or only be mentioned in the documentation).

kleini:
Will try to start from scratch with those minimal configuration options and document, what needs to be added to get useful optimization results. I will start with predictions first.

Issue #806, opened 2025-12-11
[BUG]: Battery 100% but still wants to charge from grid


a8291:
### Describe the issue:

The last few days i get some weird calculations. Basically it just wants to charge all day from grid or suggest to avoid discharge from battery. I am using EOS Connect  0.2.28 and also tried the developer version.  

I also observed over the last few weeks that EOS wants to charge more then there is physical capacity.

<img width="2433" height="1603" alt="Image" src="https://github.com/user-attachments/assets/8f524986-6d3c-47f1-aaaf-bb8373fcb1b3" />

### Reproduceable code example:

```python

```

### Error message:

```shell

```

### Version information:

EOS runs as docker in the newest version using  the HA EOS local build add-on.

a8291:
Forgot to add the battery settings...

<img width="1276" height="1493" alt="Image" src="https://github.com/user-attachments/assets/25e4f222-30a9-4856-b0e1-f1f7408e9693" />

a8291:
After deleting the HA EOS local build add-on + data it seems to work for now.

a8291:
Ok seems i was to fast. Now we are back at the same situation.

<img width="2422" height="1606" alt="Image" src="https://github.com/user-attachments/assets/88a543d5-6b9c-4b51-8a95-325da7bb8023" />

janosch1337:
<img width="1874" height="976" alt="Image" src="https://github.com/user-attachments/assets/07e24e60-c249-422c-9a49-2eb990592a57" />

same problem here

a8291:
Any update on this? Is there anything further on log side you may need to debug this?

b0661:
> Any update on this? Is there anything further on log side you may need to debug this?

As you are using EOS connect we can not do any trouble shooting. EOS connect does it's own data setup for optimization. You may analyse the REST call that EOS connect directs to the EOS /optimize endpoint to see what is happening.

MatzeBu:
I think there is nearly the same issue with discharge. 
There is a min soc in the configuration. But the calculation says sometimes discharge allowed, also when the soc is less than the min soc.

b0661:
> There is a min soc in the configuration. But the calculation says sometimes discharge allowed, also when the soc is less than the min soc.

I just checked, and there maybe a subtle bug within the PV inverter simulation and the battery handling in the overall simulation. In this special case battery simulation feedback within the PV inverter simulation might not update correctly the overall charge/ discharge planning.

[...]

PR #914, opened 2026-02-27
Fix genetic optimizer: charge rates, SOC accuracy, and AC-charge scheduling

cnadler86:
### Use charge_rates for charge power candidates (dbddec1)
bat_possible_charge_values used a hardcoded range instead of the battery's configured charge_rates, producing schedules with physically impossible charge levels. Now derives candidates directly from battery.charge_rates.

### Record SOC at start of interval (9f4410e)
SOC was snapshotted after the inverter call, causing a one-interval lag in reported values. Fixed by saving SOC before the inverter is called.

### Account for future PV surplus in AC-charge penalty (058425d)
The break-even penalty in evaluate() only counted the initial SOC as "free energy", ignoring future PV that would fill the battery for free. This caused early AC charging that displaced solar generation. Fixed by pre-computing a per-hour suffix sum of storable PV energy and adding it to free_ac_wh for each candidate charge hour.

### Separate raw gene values from SOC-clamped operation factors (d862439)
battery1_*_op_factor previously exposed raw gene values, which could command physically impossible charges into a nearly-full battery. A new _soc_clamped_operation_factors() method now scales AC charge, DC charge, and discharge factors according to actual SOC headroom. Raw values are preserved in genetic_*_factor columns for diagnostics.

### Update test fixtures (5c4a867)
All four optimize_result_*.json fixtures regenerated to match the updated algorithm. Improved Gesamtbilanz_Euro values confirm the scheduler now makes better use of available PV.

### DIFF
--- a/src/akkudoktoreos/optimization/genetic/geneticsolution.py
+++ b/src/akkudoktoreos/optimization/genetic/geneticsolution.py
@@ -305,7 +305,7 @@ class GeneticSolution(ConfigMixin, GeneticParametersBaseModel):
                 ac_to_dc_eff = float(inv_list[0].ac_to_dc_efficiency) if inv_list else 1.0
                 max_ac_cp_w = (
                     float(inv_list[0].max_ac_charge_power_w)
-                    if inv_list
+                    if inv_list and inv_list[0].max_ac_charge_power_w is not None
                     else float(bat.max_charge_power_w)
                 )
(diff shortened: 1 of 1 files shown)


cnadler86:
This PR possibly fixes #798, #806 and #844.

NormannK (approved):

cnadler86:
I opened a new PR, since when I looked again in the code I saw that the whole accounting of future PV energy will probably need a bigger refactoring. So I am splitting the PR into smaller changes (#926). I think we could close this one and continue in the new one.

PR #926, opened 2026-03-04
Fix genetic optimizer: charge rates, SOC accuracy, and minor bugfixes

cnadler86:
This is a replacement of #914 in order to introduce smaller changes.

### Record SOC at start of interval (3e0ff5495ece028d884f4faaca9fa742775c44b1)
SOC was snapshotted after the inverter call, causing a one-interval lag in reported values. Fixed by saving SOC before the inverter is called.

### Use charge_rates for charge power candidates (46002ef4bff7ff299c8ec14352097049f66f2b6c)
bat_possible_charge_values used a hardcoded range instead of the battery's configured charge_rates, producing schedules with wrong charge levels. Now derives candidates directly from battery.charge_rates.

### Separate raw gene values from SOC-clamped operation factors (667f7e06199549c47e24d8629c3aa691d4e366db)
battery1_*_op_factor previously exposed raw gene values, which could command physically impossible charges into a nearly-full battery. A new _soc_clamped_operation_factors() method now scales AC charge, DC charge, and discharge factors according to actual SOC headroom. Raw values are preserved in genetic_*_factor columns for diagnostics.

### Update test fixtures (5ab155dd6f46282f54db2e60942fb7ea57af7ae9)
All four optimize_result_*.json fixtures regenerated to match the updated algorithm. Improved Gesamtbilanz_Euro values confirm the scheduler now makes better use of available PV.

## Comparing test results
Here is a comparison about the result before updating the test fixtures:

```text
FAILED tests/test_geneticoptimize.py::test_optimize[optimize_input_1.json-optimize_result_1.json-3-0] - assert 1.053854835771092 == 1.4988843060949502 ± 1.5e-06
FAILED tests/test_geneticoptimize.py::test_optimize[optimize_input_2.json-optimize_result_2.json-3-0] - assert 12.78299149726557 == 12.490712506647688 ± 1.2e-05
FAILED tests/test_geneticoptimize.py::test_optimize[optimize_input_1.json-optimize_result_1_be.json-3-1] - assert 0.9004618481798127 == 0.5201969277249601 ± 5.2e-07
FAILED tests/test_geneticoptimize.py::test_optimize[optimize_input_2.json-optimize_result_2_be.json-3-1] - assert 14.366070195145795 == 11.91553407959862 ± 1.2e-05
```

It is true that this update produces worst results in 3/4 test. This needs further investigation.

## Fixes
[45af18b](https://github.com/Akkudoktor-EOS/EOS/pull/926/commits/45af18b81ae2270cbeb23816d6a5ae621dee98c8): home appliance will now start also at 00:00
[22ce43e](https://github.com/Akkudoktor-EOS/EOS/pull/926/commits/22ce43eb379a101300adcd3fe2829dfb3b904241): invalidate all indexes right
[ac89cb5](https://github.com/Akkudoktor-EOS/EOS/pull/926/commits/ac89cb514f8cc474878dd5a3f3f0a5eb94b18e83): remove some double code


cnadler86:
I messed up with the output of the tests (ubterstood it the wrong way). The algorithm is performing worst now, at least when looking at the numbers. will take closer look at the changes, but at least on what I saw, the solution plots were better (at least the ac charge rates and soc line were plausible).

b0661:
@cnadler86 ,  replaced by #949

Issue #919, opened 2026-02-28
[BUG]: Predictions dashboard can not be opened (Timezone mismatch)


cp-fabian-pittroff:
### Describe the issue:

The dashboard does not open and produces a HTTP 500 error.

Workaround: Set container/server timezone the same as the config one.

```yaml
environment:
  TZ: "Europe/Berlin" #same as config
```

### Reproduceable code example:

```python
Server running with timezone e.g. `UTC` and configured timezone `Europe/Berlin`.

config.json
{
  "general": {
    "timezone": "Europe/Berlin",
  }
}
```

### Error message:

`ValueError: Timezone mismatch: tz='Europe/Berlin' but data uses 'UTC'`

<details><summary>Logs / Stacktrace</summary>
<p>

```shell
INFO:     127.0.0.1:49856 - "GET /v1/prediction/dataframe?keys=pvforecast_ac_power&keys=elecprice_marketprice_kwh&keys=weather_relative_humidity&keys=weather_temp_air&keys=weather_ghi&keys=weather_d
ni&keys=weather_dhi&keys=loadforecast_power_w&keys=loadakkudoktor_std_power_w&keys=loadakkudoktor_mean_power_w HTTP/1.1" 500 Internal Server Error
Exception in ASGI application                                                                                                                                                                         
Traceback (most recent call last):                                                                                                                                                                    
  File "/opt/venv/lib/python3.13/site-packages/uvicorn/protocols/http/httptools_impl.py", line 416, in run_asgi                                                                                       
    result = await app(  # type: ignore[func-returns-value]                                                                                                                                           
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                           
        self.scope, self.receive, self.send                                                                                                                                                           
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                           
    )                                                                                                                                                                                                 
    ^                                                                                                                                                                                                 
  File "/opt/venv/lib/python3.13/site-packages/uvicorn/middleware/proxy_headers.py", line 60, in __call__                                                                                             
    return await self.app(scope, receive, send)                                                                                                                                                       
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                       
  File "/opt/venv/lib/python3.13/site-packages/fastapi/applications.py", line 1160, in __call__                                                                                                       
[...]

PR #927, opened 2026-03-05
fix: pydantic datetime dataframe timezone conversion

b0661:
Closes: #919

Waits for PR #925 to be merged.


NormannK:
rdy to be updated

PR #949, opened 2026-03-14
fix: genetic optimizer charge rates, SoC accuracy, and minor bugfixes

b0661:
This is @cnadler86 's pull request #926, just rebased to see the changes clearly and understand the test errors:
-> tests/testdata/optimize_result_2_full.json adaptation was missing

With new version:
- total costs increased by 2.3 %
- total losses increased by 0.2 %

What has changed:

* record battery SOC at the start of the interval for accurate display

* map battery charge rates to GeneticOptimizationParameters and update related logic

  Instead of using the EV charge rates, the battery now has its own charge rate defined in the GeneticOptimizationParameters to better separate those entities.

* separate raw gene values from SOC-clamped op factors

  The genetic_*_factor columns in the OptimizationSolution dataframe now always carry the raw gene values (optimizer intent), while the battery1_*_op_mode / battery1_*_op_factor columns and FRBCInstruction operation_mode_factor reflect SOC-clamped effective values that can actually be executed given the battery's state of charge at each hour.

  Adds GeneticSolution._soc_clamped_operation_factors():
    - AC charge factor: proportionally scaled down when battery headroom (max_soc - current_soc) is less than what the commanded factor would store in one hour; zeroed when battery is full.
    - DC charge factor: zeroed when battery is at or above max SOC.
    - Discharge: blocked when SOC is at or below min SOC.

  Both optimization_solution() and energy_management_plan() pass the
  clamped values to _battery_operation_from_solution(), so the plan
  instructions the HEMS receives reflect physically achievable targets.

* ensure max AC charge power is only used if defined

* update fixtures for PV suffix + SOC-clamp algorithm changes

* handle None case for home appliance start hour in simulation

  if the hour is zero, the the home appliance wont start

* update handling of invalid charge indices in autocharge hours

  Seems to be a bug, since all invalid indexes needs to be invalidated (also the first one)

* remove double code in simulation preparation and improve comments

* update version

Created by: @cnadler


b0661:
@drbacke , @NormannK , @cnadler86 are these changes ok? Shall we merge that, given that there is an improvement in accuracy of ac_charge_factor, dc_charge_factor and discharge_allowed.

The deviations compared to the current solution (-) are:

Full run:

```text
        - "Gesamtbilanz_Euro": 7.648978330847156,
        + "Gesamtbilanz_Euro": 7.824605715847156,
        - "Gesamt_Verluste": 7633.823819992856,
        + "Gesamt_Verluste": 7647.623819992857,
        -+ "Gesamteinnahmen_Euro": 0.0,
        - "Gesamtkosten_Euro": 7.648978330847156,
        + "Gesamtkosten_Euro": 7.824605715847156,
``` 

Limited run 1:

```text
        - "Gesamtbilanz_Euro": 0.5201969277249601,
        + "Gesamtbilanz_Euro": 0.9004618481798127,
        - "Gesamt_Verluste": 3181.0071533862283,
        + "Gesamt_Verluste": 2804.7326610375394,
        - "Gesamteinnahmen_Euro": 0.8774861504302851,
        + "Gesamteinnahmen_Euro": 0.9524762008447317,
        - "Gesamtkosten_Euro": 1.3976830781552452,
        + "Gesamtkosten_Euro": 1.8529380490245444,
```

Limited run 2:

```text
        - "Gesamtbilanz_Euro": 12.490712506647688,
        + "Gesamtbilanz_Euro": 12.78299149726557,
        - "Gesamt_Verluste": 8517.353033733361,
        + "Gesamt_Verluste": 5766.286846118221,
        - "Gesamteinnahmen_Euro": 0.0024471948828359577,
        + "Gesamteinnahmen_Euro": 0.5042016086902498,
        - "Gesamtkosten_Euro": 12.493159701530523,
        + "Gesamtkosten_Euro": 13.28719310595582,
```

cnadler86:
You can merge then, if you want. Haven't investigate why the results are a little worse, since I am rewriting the whole Simulation and optimization framework more or less from scratch (that will be a major change)

b0661:
> You can merge then, if you want. Haven't investigate why the results are a little worse, since I am rewriting the whole Simulation and optimization framework more or less from scratch (that will be a major change)

If you do that, please create a new algorithm with a new algorithm name and clearly separate the sources. The algorithm shall become select-able by the configuration. Please do not change the existing GENETIC algorithm. As I am experimenting also with another GA algorithm (see [here](https://github.com/b0661/EOS/tree/feat/genetic2)) please do not use GENETIC2 for your algorithm name.

NormannK:
its hard to say which solution is more right or "optimal". 
quite a while ago we circulated the idea of having some tests where the results are so easy that we (human) can understand them. 
if we have multiple algorithm that would be something to get at least a basic information if all work the intended way.

drbacke:
> You can merge then, if you want. Haven't investigate why the results are a little worse, since I am rewriting the whole Simulation and optimization framework more or less from scratch (that will be a major change)

First review: Cool ideas😊 I think the Simulation/Opti is more realistic with this pr.

PR #1124, opened 2026-06-28
Support a 15-minute optimization interval (genetic optimizer + native 15-min Tibber prices)

chloepriceless:
Builds on the new forecast providers introduced in `codex/fehler-im-ladealgorithmus-untersuchen` to bring end-to-end 15-minute operation to EOS.

**Retargeted** from `main` to `codex/fehler-im-ladealgorithmus-untersuchen`, because the 15-minute grid only delivers real value together with the sub-hourly providers added on that branch (pvnode, Forecast.Solar, Solcast, Tibber).

### What this adds
- **15-minute optimization interval for the genetic optimizer.** `optimization.interval` now accepts `900` (15 min) in addition to the default `3600` (1 h). The genetic optimizer, device power caps (battery/inverter/home appliance) and the solution/plan serializers are slot-aware (`total_slots = prediction.hours * 3600/interval`). The default `3600 s` keeps byte-identical hourly behaviour.
- **Central resampling, not per-provider rebuilds.** Provider records stay at their native resolution; `key_to_array(interval=…)` maps them onto the slot grid (PV linear-interpolated, prices/load ffill, power series scaled to Wh/slot). The new sub-hourly PV providers (pvnode 15 min, Solcast 30 min, Forecast.Solar) feed their native resolution straight into the quarter-hour grid.
- **Native 15-minute Tibber prices.** `ElecPriceTibber` now requests `priceInfoRange(resolution: QUARTER_HOURLY)` and stores prices at their native resolution instead of averaging them to 1 h, so both the hourly and the 15-minute optimizer get the correct grid. The seasonal ETS extrapolation is resolution-agnostic and stays identical at the default hourly resolution.

### Tests
- `tests/test_optimization_interval.py` → 4 passed (end-to-end 192-slot day, 900 s output cadence)
- `tests/test_elecpricetibber.py` → 15 passed (native-15min preservation + hourly backward-compat)

### Note on pre-existing failures
`tests/test_geneticoptimize.py` has 4 failures that are **pre-existing on this codex branch** (commit `cd5cdf8` "fix: limit EV charging to unmet min SoC" changed `eautocharge_hours_float` without updating the expected `optimize_result_*.json` fixtures). They are inherited from the base branch, not introduced by this PR — verified by a three-way comparison of both merge parents.

PR #1140, opened 2026-07-09
elecprice(Tibber): serve native 15-minute (QUARTER_HOURLY) prices

chloepriceless:
This PR builds on the Tibber price extrapolation work on this branch and makes the Tibber provider deliver **native 15-minute exchange prices**.

## What & why
EOS already resamples stored prediction records onto the optimization grid on demand (`key_to_array(interval=optimization.interval)`). The correct place to support the quarter-hour grid is therefore the provider: it should store its **native** resolution instead of pre-averaging to hourly. Then both the hourly (`interval=3600`) and the 15-minute (`interval=900`) optimizer are fed the correct grid automatically.

## Changes
- **GraphQL query**: `priceInfoRange(resolution: HOURLY, last: 840)` → `resolution: QUARTER_HOURLY, last: 960`. `priceInfo.today`/`tomorrow` are unchanged (they already return the finer resolution). Tibber has offered `QUARTER_HOURLY` in DE since 1 Oct 2025.
- **`_hourly_series` → `_normalize_series`**: dedupe by timestamp (mean) + sort, **without** 1h aggregation, so a 15-minute API resolution stays 15-minute. New `_resolution_seconds` helper infers the native slot size from the median of timestamp diffs (robust fallback 3600 s).
- **Resolution-agnostic ETS extrapolation**: seasonal windows and history thresholds are scaled by `slots_per_hour` (`168/24 * slots_per_hour` seasonal periods; needed-forecast length and the prediction index step are computed in slots). At the default hourly resolution (`slots_per_hour == 1`) the behaviour is **identical** to before.

## Tests
- Replaced the old 1h-averaging test with a resolution-preserving test (+ a dedup test).
- Asserted the query now contains `QUARTER_HOURLY`.
- Added a 15-minute end-to-end test: native storage stays 15-min, slot-based seasonal periods (daily-only history → 96), 15-minute forecast index.
- The existing hourly extrapolation tests stay green unchanged (backward-compatibility proof).

`uv run python -m pytest tests/test_elecpricetibber.py -q` → **15 passed**.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
   - `PVForecastPVNode` — native 15-minute forecasts from the pvnode.com API.
   - `PVForecastForecastSolar` — forecasts from the free Forecast.Solar API.
   - `PVForecastSolcast` — forecasts from the Solcast rooftop-site API.
+- The Tibber electricity price provider now requests native 15-minute exchange prices
+  (`priceInfoRange(resolution: QUARTER_HOURLY)`) and stores them at their native
+  resolution instead of pre-averaging to hourly values, so EOS can resample them onto
+  either the hourly or the quarter-hour optimization grid. The seasonal price
+  extrapolation is resolution-agnostic and stays identical at the default hourly
+  resolution.
 
 ## 0.3.0 (2026-03-17)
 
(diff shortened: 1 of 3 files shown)


chloepriceless:
Superseded by #1124, which now targets this codex branch and carries the full 15-minute change (optimizer + native-15min Tibber together). Closing this Tibber-only PR to avoid duplication.

PR #1157, opened 2026-07-17
Direct marketing: overnight reserve (price-aware) + /v1/prediction/import serialization fix

chloepriceless:
Two additions from our production fork (running on a 43 kWh Victron system), rebased onto this branch:

## 1. `fix(server)`: /v1/prediction/import 400 on DataFrame bodies

A PUT body that pydantic validates as `PydanticDateTimeDataFrame` reached `json.dumps(data)`, which cannot serialize BaseModel instances — every such import failed with HTTP 400 *"Object of type PydanticDateTimeDataFrame is not JSON serializable"* (reproduced on this branch yesterday). One-liner: serialize models via `model_dump_json()`, keep `json.dumps` for plain dicts.

## 2. `feat(optimization)`: overnight reserve for direct marketing

Without a reserve, `battery_grid_export` sells the pack into the evening prices and re-imports the whole night's household load at the (usually higher) import price. We measured this yesterday by replaying our production inputs against this branch: the remaining battery was sold at 18.9 ct/kWh in the first evening slot while the night re-import cost 26.9 ct/kWh.

The reserve is a per-slot **export cap** (never a self-consumption limit): `margin × Σ max(load − PV, 0)` from each slot until PV next covers load, walked backwards. The pack rides the night on self-consumption and arrives near the floor as morning PV takes over.

**Price-aware release** (opt-in, `EOS_RESERVE_PRICE_AWARE=1`): a slot's reserve is released down to a load-adaptive safety floor when that slot's *own* export revenue beats the highest avoided night-import price of its night window by a relative margin (default 20%). Genuine price peaks still get sold; sub-threshold slots can't nibble the reserve away. Per-slot (not window-peak) on purpose — we shipped the window-peak variant first and watched it drain the reserve through sub-threshold neighbour slots.

Implementation notes:
- Pure function of the forecast arrays, computed once per `prepare()` → deterministic, plays nice with your fitness cache.
- Env-gated with behavior-preserving default for the price-aware part; the energy-balance reserve defaults ON (that default is the bugfix — happy to flip it or lift everything into `feedintariff`/`optimization` config fields if you prefer).
- 6 unit tests (`tests/test_overnight_reserve.py`); adjacent suites (inverter, genetic simulation, seeding, 15-min interval) stay green. The 4 `test_geneticoptimize` reference tests fail identically with and without this change (time-of-day dependent "Start hour not synced").

Running in production in our fork since 2026-06-18 (price-aware since 2026-06-29) — the two-pool behavior (free surplus sales + protected night reserve) is what our direct-marketing setup depends on.

### DIFF
--- a/src/akkudoktoreos/devices/genetic/inverter.py
+++ b/src/akkudoktoreos/devices/genetic/inverter.py
@@ -51,6 +51,7 @@ class Inverter:
         consumption: float,
         hour: int,
         allow_battery_grid_export: bool = False,
+        export_reserve_ac_wh: float = 0.0,
     ) -> tuple[float, float, float, float]:
         """Process one slot using probabilistic direct PV-to-load overlap.
 
@@ -59,6 +60,13 @@ class Inverter:
         PV-to-load power. The remaining load and PV surplus are then handled
         independently, because both can occur during different sub-intervals of
         the same hourly or 15-minute slot.
+
+        ``export_reserve_ac_wh`` is delivered-AC energy the battery must keep
+        for overnight household self-consumption. It only caps the
+        battery-to-grid export path (direct marketing): covering the slot's own
+        load may still use the reserve, so the pack rides the night on
+        self-consumption instead of selling out in the evening and re-importing
+        the whole night from the grid.
         """
         losses = 0.0
         grid_export = 0.0
@@ -126,6 +134,10 @@ class Inverter:
             remaining_battery_ac = (
                 self.battery.remaining_discharge_energy_wh(hour) * self.dc_to_ac_efficiency
             )
+            # Overnight reserve: energy earmarked for tonight's household load
+            # is not exportable. Self-consumption above already had full access.
+            if export_reserve_ac_wh > 0.0:
+                remaining_battery_ac = max(remaining_battery_ac - export_reserve_ac_wh, 0.0)
             export_capacity = min(remaining_inverter_ac_capacity, remaining_battery_ac)
             battery_export_ac, battery_export_losses = self._discharge_battery_to_ac(
                 export_capacity, hour
--- a/src/akkudoktoreos/devices/genetic/inverter.py
+++ b/src/akkudoktoreos/devices/genetic/inverter.py
@@ -136,8 +136,22 @@ class Inverter:
             )
             # Overnight reserve: energy earmarked for tonight's household load
             # is not exportable. Self-consumption above already had full access.
+            # The reserve caps the SOC POOL, not the per-slot power budget: a
+            # night reserve larger than one slot's discharge budget must NOT
+            # zero out export in every slot — only the energy above
+            # min_soc + reserve is sellable, at full slot power.
             if export_reserve_ac_wh > 0.0:
-                remaining_battery_ac = max(remaining_battery_ac - export_reserve_ac_wh, 0.0)
+                disch_eff = self.battery.discharging_efficiency
+                if self.dc_to_ac_efficiency > 0.0 and disch_eff > 0.0:
+                    reserve_dc_wh = export_reserve_ac_wh / (self.dc_to_ac_efficiency * disch_eff)
+                else:
+                    reserve_dc_wh = export_reserve_ac_wh
+                soc_exportable_ac = (
+                    max(self.battery.soc_wh - self.battery.min_soc_wh - reserve_dc_wh, 0.0)
+                    * disch_eff
+                    * self.dc_to_ac_efficiency
+                )
+                remaining_battery_ac = min(remaining_battery_ac, soc_exportable_ac)
             export_capacity = min(remaining_inverter_ac_capacity, remaining_battery_ac)
             battery_export_ac, battery_export_losses = self._discharge_battery_to_ac(
                 export_capacity, hour
--- a/src/akkudoktoreos/optimization/genetic/genetic.py
+++ b/src/akkudoktoreos/optimization/genetic/genetic.py
@@ -1,5 +1,6 @@
 """Genetic algorithm."""
 
+import os
 import random
 import time
 from collections import OrderedDict, defaultdict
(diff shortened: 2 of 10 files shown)


chloepriceless:
## Follow-up: measurement data from replaying our production inputs against this branch

We replayed our production system's exact EOS inputs (15-min series for price/feed-in/load/PV, 43 kWh battery, eff 0.922, LCOS 0.036, start SoC 14%, `direct_marketing_enabled=true`, GA 300x400) against this branch — twice: vanilla yesterday, and today with this PR's reserve applied. Two findings you may find useful for the live testing:

### 1. The reserve part of this PR works as intended
Vanilla branch sold the remaining 14% SoC in the first evening slot @ 18.9 ct/kWh and re-imported the night @ 26.9 ct/kWh. With the reserve, the pack rides the night on self-consumption (14% → 5% by morning), sub-threshold battery sales drop from the full remainder to 0.5 kWh.

### 2. The EMS-produced plans never charge the battery from PV — but your fitness model actually prefers charging
In all our EMS runs the best individual has `dc_charge = 0` in **every** slot: SoC stays flat at min_soc all day despite negative midday prices and an 18 ct evening peak, PV is exported directly even at −0.03 ct (booked as losses).

To separate model from search, we evaluated a hand-built individual under your own `GeneticSimulation.simulate()` (same arrays, reserve disabled):

| Variant | Gesamtbilanz (costs − revenue, lower = better) |
|---|---|
| EMS best (dc=0, no export) | **+2.54 EUR** |
| hand-built: dc-charge in PV-surplus slots + export in evening slots > 16 ct | **−1.74 EUR** |

So the model rewards charging by ~4.3 EUR/day for this profile — the GA just doesn't reach that region. Two narrower observations:

- A **fresh** in-process run (no warm start, `fixed_seed=42`, 120x80) *does* charge (dc active in 103/192 slots, SoC max 100%) — but ends with **0 export slots** and +2.03 EUR. The educated-guess export candidates (export-quantile family) exist, so something later in the pipeline (possibly the export→self-consumption local search shifts, `_grid_export_shift_candidates`) appears to strip coordinated export genes.
- The **warm-started** EMS runs are worse than fresh runs: once a degenerate previous solution exists, the warm-start neighbourhood seems to keep the population in the never-charge basin across runs.

FWIW, we hit the same class of problem in our fork earlier and solved it with a deterministic greedy seed (charge PV surplus, sell best-revenue slots, SoC-simulated so it stays feasible) that anchors the population. Happy to share the replay harness (prod series + config mirror + per-slot diff) if useful — it reproduces this in ~10 minutes.

chloepriceless:
## Correction + fix pushed: the "GA never charges" diagnosis was largely OUR measurement bug — retracted

In fairness to your GA, two contaminations invalidated most of my previous comment:

1. **Yesterday's vanilla replay carried a phantom EV** (50 kWh, min_soc 70%, `ev_soc_miss` penalty, no SoC measurement) accidentally mirrored from our production config — that alone skews any run.
2. **Today's runs hit a real bug in THIS PR's reserve port** (my code, not yours): the export cap subtracted the overnight reserve from `remaining_discharge_energy_wh()`, which is already capped to ONE slot's discharge power budget (4.5 kWh at 15 min). Any night reserve larger than that budget zeroed battery export in EVERY slot → direct marketing silently disabled → charging became pointless (stranded energy, zero terminal value) → the GA *correctly* converged to `dc_charge = 0`. Your search wasn't failing to find charging — our cap had made charging economically worthless.

**Fixed in 8e61295**: the reserve now caps the SOC pool (energy above min_soc + reserve, at full slot power), matching the semantics we run in production. Regression test included (`test_reserve_caps_soc_pool_not_slot_power_budget`).

### After the fix — your GA does exactly the right thing

Clean setup (no EV, `fixed_seed=42`, only 120×80 — a fraction of the default budget), same 43 kWh replay day:

| | before fix | after fix |
|---|---|---|
| best fitness (costs − revenue, lower better) | −0.11 | **−3.60** |
| best educated export seed | +1.49 | **−1.91** |
| behavior | never exports, barely charges | charges midday (58 dc slots), exports **~24 kWh** into the evening peak at full slot power, holds **13% SoC** for the night |

So with export actually possible, model AND search converge to the intended two-pool direct-marketing behavior. I'm retracting the `_grid_export_shift_candidates` suspicion and the "probability table" framing from my previous comment.

Still standing from the earlier report: the `/v1/prediction/import` serialization fix, and the reserve rationale itself — without a reserve the remaining pack is sold into the first evening slot and the whole night is re-imported at the higher import tariff.

chloepriceless:
Two updates on this branch today, from validating the direct-marketing path end-to-end against a live server instance (prod-mirrored inputs, 192 15-min slots):

**1. f2faaf5 `fix(optimization): direct marketing must use the imported feed-in revenue series` — now server-validated.** Before this fix, `GeneticOptimizationParameters.prepare()` silently replaced the imported feed-in series with `elecprice_marketprice_wh` whenever the configured provider was `FeedInTariffImport` (missing from `MARKET_PRICE_FEED_IN_TARIFF_PROVIDERS`). For bridges that push the resolved end-customer *import* price to `ElecPriceImport` (flat fixed tariff in our case), the GA then saw feed-in == import price in every slot — a world where battery arbitrage can never pay — and correctly converged to never cycling the battery. That is the root cause of the degenerated server EMS runs reported earlier in this thread; in-process replays bypass `prepare()`, which is why they always converged fine. Validated with temporary instrumentation in `optimierung_ems`: feed-in reached the GA flat at 0.000269 EUR/Wh (= import price) before the fix, and as the real variable market series (max 0.000225 EUR/Wh, negatives preserved) after it.

**2. 370c14c `feat(prediction): FeedInTariffDvhubOnline provider` — went in along the way and is now part of this PR.** A market-revenue feed-in provider like EnergyCharts/Tibber: fetches EPEX day-ahead prices (15-min slots, EUR/MWh) from the public dvhub.online price API (`GET /api/prices?start=YYYY-MM-DD&end=YYYY-MM-DD&zone=DE-LU`, configurable `base_url`/`zone`) and stores the raw market price as `feed_in_tariff_wh` — no import charges or VAT. Registered in `MARKET_PRICE_FEED_IN_TARIFF_PROVIDERS` so direct marketing uses its series directly. Comes with 5 unit tests plus an opt-in live smoke test (`EOS_DVHUB_ONLINE_LIVE=1`), all passing.

chloepriceless:
## Root cause of the degenerate EMS plans found + fixed: direct marketing silently replaced the imported feed-in revenues with the electricity price

We instrumented `optimierung_ems` on our replay instance and dumped the parameters the GA actually receives. Smoking gun:

```
TEMPDIAG: feedin_max=0.000269  feedin_nonzero=192   # flat 26.9 ct/kWh in ALL 192 slots = our END-CUSTOMER IMPORT price
```

**Cause** (`geneticparams.py`): with `direct_marketing_enabled`, `prepare_optimization_parameters` substitutes `elecprice_marketprice_wh` for the feed-in series whenever the configured provider is not in `MARKET_PRICE_FEED_IN_TARIFF_PROVIDERS` — and **`FeedInTariffImport` was missing from that set**. For your setups (elecprice = spot) the substitution is a reasonable approximation. External EMS bridges like ours, however, push the resolved end-customer import price into elecprice (needed for import-cost math) and the per-slot market revenue into `FeedInTariffImport` — so the GA saw *feed-in == import price* in every slot: a world in which battery arbitrage can never pay. Its "degenerate" never-cycle plans were the **correct optimum for the wrong world**. The GA was blameless the whole time.

**Fixed in f2faaf5**: `FeedInTariffImport` added to the market-price set, the `except`-fallback now logs a warning instead of silently rewriting the economics, plus regression tests.

**End-to-end proof** (same replay setup, EMS path, fresh data): parameter dump now shows the real market series (`feedin_max=0.000225`, negative midday), and the EMS plan cycles the battery properly — SoC 98% → 62% through evening sales → 25% overnight self-consumption → recharges next midday; zero grid import over the 48 h window.

The branch also gained a `FeedInTariffDvhubOnline` provider (EPEX day-ahead from dvhub.online) as an additional market-revenue source (370c14c).

Diagnostic lesson we'll keep: when an optimizer degenerates, dump the parameters it actually receives before suspecting the search — we burned five wrong hypotheses (env, charge_rates, data staleness, window length, seeding) before looking at the inputs.

chloepriceless:
Follow-up: the first full server EMS run with f2faaf5 active has completed — **the degeneration is gone**. Same live instance, prod-mirrored inputs (192 15-min slots), real measured SoC (98%):

- fitness/net balance: **−16.49 €** (costs 2.71 − revenues 19.20) vs. the degenerated runs reported earlier (flat SoC, no charging, positive balance)
- the GA now **charges from PV** (26 dc-charge slots > 0) and **sells into the evening peaks** (full ~3.8 kWh slots at 18:45/19:00/20:00/22:00, SoC 98→56%)
- overnight grid import 22:00–06:00: **0.00 kWh**, with ~14% SoC held through the night (the overnight reserve of this PR doing its job)

So with the imported feed-in revenue series actually reaching the GA, both the search and the reserve behave as designed on a live server — no remaining EMS-vs-in-process delta on our side.

chloepriceless:
**Withdrawing this PR**

We ran our own tests against this branch. Your implementation handles this better than ours did, so the overnight reserve is no longer necessary and our version of it is not needed any more — we are closing this.

The two unrelated fixes that were bundled in here are now separate: #1223 and #1224. Thanks for the branch!

PR #675, opened 2025-09-10
Feature/api english translation

7tobias:
## API Field Name Standardization
  This PR translates API field names from German to English to make the API more accessible to international developers while maintaining full backward compatibility.
 
  ## Changes
 
  ### Field Name Translations
 
  Renamed 18 fields across two main classes:
 
  **GeneticEnergyManagementParameters (5 fields):**
  - `pv_prognose_wh` → `pv_forecast_wh`
  - `strompreis_euro_pro_wh` → `electricity_price_per_wh`
  - `einspeiseverguetung_euro_pro_wh` → `feed_in_tariff_per_wh`
  - `preis_euro_pro_wh_akku` → `price_per_wh_battery`
  - `gesamtlast` → `total_load`
 
  **GeneticSimulationResult (13 fields):**
  - `Last_Wh_pro_Stunde` → `load_wh_per_hour`
  - `EAuto_SoC_pro_Stunde` → `ev_soc_per_hour`
  - `Einnahmen_Euro_pro_Stunde` → `revenue_per_hour`
  - `Gesamt_Verluste` → `total_losses`
  - `Gesamtbilanz_Euro` → `total_balance`
  - `Gesamteinnahmen_Euro` → `total_revenue`
  - `Gesamtkosten_Euro` → `total_costs`
  - `Kosten_Euro_pro_Stunde` → `costs_per_hour`
  - `Netzbezug_Wh_pro_Stunde` → `grid_consumption_wh_per_hour`
  - `Netzeinspeisung_Wh_pro_Stunde` → `grid_feed_in_wh_per_hour`
  - `Verluste_Pro_Stunde` → `losses_per_hour`
  - `akku_soc_pro_stunde` → `battery_soc_per_hour`
  - `Electricity_price` → `electricity_price` (already English)
 
  ### Backward Compatibility
 
  **Input:** Both English and German field names accepted via Pydantic `AliasChoices`
  **Output:** Both field names included in JSON via `@computed_field` properties
 
  No breaking changes - existing consumers continue to work.
 
  ### Currency-Neutral Field Names
 
  Removed currency-specific references from field names (e.g., `euro`) to support any currency. The system works with EUR, GBP, USD, SEK, CHF, etc. - all monetary values use the locally configured currency.
 
  Resolves #628

PR #1164, opened 2026-07-19
feat: rename remaining German API fields and deprecate legacy endpoints

7tobias:
Follow-up to #675, as discussed there.

- Rename the remaining German input fields `pv_akku` -> `pv_battery` and `eauto` -> `ev`, and the solution fields `eautocharge_hours_float` -> `ev_charge_hours_float` and `eauto_obj` -> `ev_obj`. Same backward compatibility pattern as #675: the German names are still accepted on input and re-emitted in responses, marked deprecated in the OpenAPI schema.
- Mark the legacy endpoints `/strompreis`, `/gesamtlast` and `/gesamtlast_simple` as deprecated in the OpenAPI schema. Behavior is unchanged and no new endpoints are needed - the replacements are the existing generic `/v1/prediction/*` endpoints, as already noted in the docstrings.
- Rename the remaining eauto-prefixed internal variables and translate the last German code comments.
- Replace the last euro reference in the battery LCOS log message with the amount wording.
- State in CONTRIBUTING.md that code, comments, identifiers and API field names are written in English.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -54,6 +54,8 @@ The server can be started with `make run`. A full overview of the main shortcuts
 
 ### Code Style
 
+All code, comments, docstrings, identifiers and API field names are written in English.
+
 Our code style checks use [`pre-commit`](https://pre-commit.com).
 
 To run formatting automatically before every commit:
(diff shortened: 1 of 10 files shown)


7tobias:
One question on the CONTRIBUTING.md addition: is the English-only wording ok as stated, or would you prefer it phrased differently (or in a separate PR)?

PR #1169, opened 2026-07-19
fix: fill prediction horizon in FeedInTariffFixed

### LINKED ISSUE #1167: [BUG]: FeedInTariffFixed writes a single value at now() instead of filling the prediction horizon
### Describe the issue

With `feedintariff.provider = "FeedInTariffFixed"` and `feed_in_tariff_kwh` set, the `feed_in_tariff_wh` prediction series never covers the prediction window. It only contains isolated points whose timestamps are the exact moments the provider ran:

```
GET /v1/prediction/series?key=feed_in_tariff_wh
"2026-07-19T16:34:13.513330+02:00": 8.2e-05,
"2026-07-19T16:34:15.842...":       8.2e-05,
...   (11 points, sub-second timestamps - one per parameter-preparation retry)
```

`key_to_array()` over the parameter window (start of day + `prediction.hours`, interval grid) therefore yields no usable series, which makes `GeneticOptimizationParameters.prepare()` fail (masked by the bare `except:` — see #1166 about the misleading error message).

### Root cause

`prediction/feedintarifffixed.py::_update_data` writes exactly one record at the current time:

```python
feed_in_tariff_wh = feed_in_tariff / 1000
await self.update_value(to_datetime(), "feed_in_tariff_wh", feed_in_tariff_wh)
```

A fixed-tariff provider should populate `feed_in_tariff_wh` across the full prediction horizon (one entry per optimization interval starting at `ems_start_datetime`), analogous to `ElecPriceFixed`.

### Fix / verification

I have a fix ready on a branch (generation loop analogous to `ElecPriceFixed`, plus a regression test `test_update_data_fills_prediction_horizon`; pre-commit and pytest pass) and will open a PR referencing this issue.

### Workaround

Switch to `FeedInTariffImport` with a constant list, e.g. `import_json = {"feed_in_tariff_wh": [0.000082, ...x96]}`.

### Environment

Same as #1166 (EOS 0.3.0.dev HA add-on, image built 2026-07-18, Python 3.13, HAOS x86-64).

rockinglama:
Fixes #1167.

`FeedInTariffFixed._update_data()` wrote exactly one value at the current time on every update, so the `feed_in_tariff_wh` series never covered the prediction horizon and parameter preparation could not use it.

## Changes

- Generate one entry per optimization interval over `prediction.hours` starting at `ems_start_datetime`, analogous to `ElecPriceFixed`.
- Replace the bare `except:` with `except Exception:` (cf. #1159).
- Add regression test `test_update_data_fills_prediction_horizon`.

## Verification

- `pytest tests/test_feedintarifffixed.py`: 3 passed (new test included)
- pre-commit (isort, ruff, ruff-format, mypy): passed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01RGkbUKjgbeSCTYrS1dmwwe

### DIFF
--- a/src/akkudoktoreos/prediction/feedintarifffixed.py
+++ b/src/akkudoktoreos/prediction/feedintarifffixed.py
@@ -7,7 +7,6 @@ from pydantic import Field
 
 from akkudoktoreos.config.configabc import SettingsBaseModel
 from akkudoktoreos.prediction.feedintariffabc import FeedInTariffProvider
-from akkudoktoreos.utils.datetimeutil import to_datetime
 
 
 class FeedInTariffFixedCommonSettings(SettingsBaseModel):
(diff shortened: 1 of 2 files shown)


b0661:
> feed_in_tariff_wh series never covered the prediction horizon and parameter preparation could not use it.

Why? Parameter preparation uses forward fill ("ffill"). So a single value will become a list spaced by the given interval.

```python
                array = await cls.prediction.key_to_array(
                    key="feed_in_tariff_wh",
                    start_datetime=parameter_start_datetime,
                    end_datetime=parameter_end_datetime,
                    interval=interval,
                    fill_method="ffill",
                )
                feed_in_tariff_wh = array.tolist()
```

NormannK:
Thanks for pointing this out.

After re-checking this against the current code and the changes that landed shortly after this PR, I don't think this PR is needed anymore.

The assumption behind this change was incorrect: `feed_in_tariff_wh` is read through `key_to_array(..., fill_method="ffill")`, so a single fixed tariff value is sufficient and does not need to be stored once per optimization interval.

The optimizer failure I was seeing at the time was addressed in #1168. `key_to_array()` returns a NumPy array, but parameter preparation was calling `.to_list()` instead of `.tolist()`. That raised an `AttributeError` and was then reported as missing forecast data. #1168 fixed this for all affected arrays, including `feed_in_tariff_wh`.

The time-series/resampling code has also since been further refactored in #1205 and the `start_datetime` handling was fixed in #1214.

So keeping the interval-generation loop from this PR would now just duplicate a constant value unnecessarily.

Closing this as superseded / no longer needed.

Thanks @b0661 for catching the `ffill` behavior.

PR #1187, opened 2026-07-23
feat: add Akkudoktor feed-in provider

b0661:
This is a snippet from Andreas' development branch https://github.com/Akkudoktor-EOS/EOS/tree/feat/direct-marketing-battery-grid-export put on top of current main branch.

The `FeedInTariffAkkudoktor` provider uses raw day-ahead market prices from
`https://api.akkudoktor.net/prices` as `feed_in_tariff_wh`. It does not add electricity import
charges or VAT. Published prices are extended to the configured prediction horizon with the same
seasonal ETS or median fallback used by the Akkudoktor electricity-price provider.
    
The Akkudoktor endpoint currently forwards hourly market prices from aWATTar. With a 15-minute
optimization interval, EOS holds each hourly price constant for its four quarter-hour slots. This
keeps the slot grid consistent but does not create genuine quarter-hour market prices.


NormannK:
bug can be solved later.

PR #1190, opened 2026-07-24
chore: prepare for update of genetic algorithm

b0661:
Andreas will update the genetic algorithm for 15-minutes optimization intervals. See https://github.com/Akkudoktor-EOS/EOS/tree/feat/direct-marketing-battery-grid-export.

Copy the current GENETIC optimization algorithm to GENETIC0 to enable
to keep the algorithm with the current functionality. Also copy resources
like the load interpolator to the GENETIC0 algorithm to keep them despite
possible later changes to the interpolator.

Make the deprecated legacy /optimize endpoint use the GENETIC0 optimization
algorithm to in-fact behave the same way even if there will later be changes
to the GENETIC algorithm by Andreas. Add a new REST endpoint to provide
the unprocessed optimisation results of the GENETIC and GENETIC0 algorithm
in case one wants to use them as done with the deprecated /optimize endpoint.

Adapt the optimization configuration to have distinct configurations for the
GENETIC and the GENETIC0 algorithm.

Create a copy of the current tests for the GENETIC algorithm to be used
for the GENETIC0 algorithm. This avoids the tests for the GENETIC0
algorithm to be influenced by later changes by Andreas.


src/akkudoktoreos/core/ems.py
```diff
@@ -288,93 +320,113 @@ async def run(
...
 
-            if not genetic_parameters:
-                logger.error("Energy management run canceled. Could not prepare parameters.")
-                EnergyManagement._stage = EnergyManagementStage.IDLE
-                return
+                # --- Optimization (CPU-bound → MUST offload) ---
+                try:
+                    genetic0_optimization = Genetic0Optimization(
+                        verbose=bool(self.config.server.verbose),
+                        fixed_seed=genetic_seed,
```
NormannK: this should be
`fixed_seed=genetic0_seed`

src/akkudoktoreos/server/eos.py
```diff
@@ -1161,6 +1162,38 @@ def fastapi_energy_management_optimization_solution_get() -> OptimizationSolutio
     return solution
 
 
+@app.get("v1/energy-management/optimization/solution/{algorithm}", tags=["energy-management"])
```
NormannK: i think the / is missing

`@app.get("/v1/energy-management/optimization/solution/{algorithm}", tags=["energy-management"])`

src/akkudoktoreos/optimization/genetic0/genetic0params.py
```diff
@@ -0,0 +1,710 @@
...
+            logger.info("Prediction historic hours unknown - defaulting to 24 hours.")
+            cls.config.prediction.historic_hours = 24
+        # Check optimization definitions
+        if cls.config.optimization.horizon_hours is None:
+            logger.info("Optimization horizon unknown - defaulting to 24 hours.")
+            cls.config.optimization.horizon_hours = 24
+        if cls.config.optimization.interval is None:
+            logger.info("Optimization interval unknown - defaulting to 3600 seconds.")
+            cls.config.optimization.interval = 3600
+        if cls.config.optimization.interval != 3600:
```
NormannK: this overrides a global parameter.
if we would like to run GENETIC0 and GENETIC in one session it would reset and cause a problem.
how about a local parameter in GENETIC0?

src/akkudoktoreos/core/ems.py
```diff
@@ -288,93 +320,113 @@ async def run(
...
+                    loop = get_running_loop()
+                    start_hour = EnergyManagement._start_datetime.hour
+                    genetic0_solution = await loop.run_in_executor(
+                        None,
+                        lambda: genetic0_optimization.optimize_ems(
+                            start_hour=start_hour,
+                            parameters=cast(
+                                Genetic0OptimizationParameters, genetic0_parameters
+                            ),  # cast for mypy
+                            ngen=genetic0_individuals,
```
NormannK: why do we hand over `genetic0_individuals` for the ngen? 

individuals are not the ngen (number of generations).
it should be like this.

optimize_ems(
    parameters=genetic0_parameters,
    generations=genetic0_generations,
    individuals=genetic0_individuals,
)

maybe we just use the wrong wording for the variable?

src/akkudoktoreos/server/eos.py
b0661: Fixed

src/akkudoktoreos/optimization/genetic0/genetic0params.py
b0661: Configurations for GENETIC and GENETIC0 are now distinct.

src/akkudoktoreos/core/ems.py
b0661: This was a real bug. Now the generations configuration parameter is used (instead of individuals).

PR #1191, opened 2026-07-26
fix: root privileges drop on non unix systems

b0661:
This is a snippet from Andreas' development branch https://github.com/Akkudoktor-EOS/EOS/tree/feat/direct-marketing-battery-grid-export put on top of current main branch.

Skip root privileges drop on non Unix systems. It is just not supported.

### DIFF
--- a/src/akkudoktoreos/server/server.py
+++ b/src/akkudoktoreos/server/server.py
@@ -1,13 +1,20 @@
 """Server Module."""
 
-import grp
 import ipaddress
 import os
-import pwd
 import re
 import socket
+import sys
 import time
 from typing import Any, Optional
+
+try:
+    # Only available on Linux/Unix type systems
+    import grp
+    import pwd
+except ModuleNotFoundError:
+    grp = None  # type: ignore[assignment]
+    pwd = None  # type: ignore[assignment]
 
 import psutil
 from loguru import logger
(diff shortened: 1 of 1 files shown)


src/akkudoktoreos/server/server.py
```diff
@@ -264,6 +277,10 @@ def fix_data_directories_permissions(run_as_user: Optional[str] = None) -> None:
         run_as_user (Optional[str]): The user who should own the data directories and files.
             Defaults to current one.
     """
+    if pwd is None or not hasattr(os, "geteuid") or not hasattr(os, "chown"):
+        logger.debug("Skipping data directory ownership fix on `{sys.platform}`.")
```
NormannK: {sys.platform} will not be written this way into the line. f missing
`logger.debug(f"Skipping data directory ownership fix on `{sys.platform}`.")`

src/akkudoktoreos/server/server.py
```diff
@@ -185,6 +192,12 @@ def drop_root_privileges(run_as_user: Optional[str] = None) -> bool:
         - The target user must exist inside the container (valid entry in
           ``/etc/passwd`` and ``/etc/group``).
     """
+    if pwd is None or grp is None or not hasattr(os, "geteuid"):
+        if run_as_user is not None:
+            logger.error("Privilege switching is not supported on `{sys.platform}`.")
```
NormannK: {sys.platform} will not be written this way into the line. f missing
`logger.error(f"Privilege switching is not supported on `{sys.platform}`.")`

NormannK (changes requested):

src/akkudoktoreos/server/server.py
b0661: fixed

src/akkudoktoreos/server/server.py
b0661: fixed

NormannK (approved):

Issue #1192, opened 2026-07-26
[ENH]: Integrate direct marketing and 15-minutes interval optimization


b0661:
### Describe the enhancement or feature request:

This is to track the integration of Andreas' development branch https://github.com/Akkudoktor-EOS/EOS/tree/feat/direct-marketing-battery-grid-export into the main branch.

### Link to discussion and related issues

Pull requests that integrate Andreas' development branch into main:

- #1242
- #1195 
- #1193 
- #1189 
- #1187 
- #1165 
- #1160 
- #1154 
- #1150

Pull requests that shall be added but do not stem from Andreas' development  branch:

- #1256 
- #1235
- #1214 
- #1212 partly superseeded by 1235
- #1208 
- #1205 
- #1191 
- #1190 

Known bugs of the integration:

- #1188
- #1181
- #1204
- #1206
- #1227 
- #1254
- #1255 

### Proposed implementation

Tasks for transfer from Andreas' development branch to main:

- [x] New pv forcast providers
- [x] New electricity price providers
- [x] New feed in tariff providers
- [x] Add ElecPriceSMARD and FeedInTariffSMARD
- [x] Move static network fees configuration and management to new ElectricityFeeProvider and adapt network fee usage in other providers
- [x] Adaptions to ElecPriceEnergyCharts electricity provider and updates to FeedInTariffEnergyCharts
- [ ] Adaptions to the GENETIC optimization algorithm/ devices/ tests
- [ ] Adaptations to the GENETIC PDF visualization
- [x] Adapt deprecated /strompreis endpoint to aggregate 15-minute spot prices to hourly means.
- [ ] Fill changelog
- [ ] Update version to 0.4
- [x] Bug fix: Skip root privilege drop on non Unix systems
- [x] Bug fix: Use model_dump_json on import_json for pydantic models 
- [x] Bug fix: Windows compliant path check in version calculation

Tasks for design changes due to changes by Andreas'

- [x] Copy current GENETIC optimization algorithm/ devices/ tests and provide as extra GENETIC0 algorithm/ devices/ tests for legacy use 
- [x] Change PDF visualization to be created on demand and per optimization algorithm
- [x] Generate a new ElectricityFeeProvider class of providers for any fees that apply to electricity prices and feed-in tariffs. 
- [ ] Adapt configuration to be usable for the changed GENETIC algorithm and the changed simulation devices and also for the existing GENETIC0 algorithm and the related devices (after copy from GENETIC).

PR #1193, opened 2026-07-26
feat: add dvhubonline feed-in tariff provider

b0661:
This is a snippet from  @chloepriceless' pull request #1192 to Andreas' development branch https://github.com/Akkudoktor-EOS/EOS/tree/feat/direct-marketing-battery-grid-export put on top of current main branch.
    
Fetches GET /api/prices?start&end&zone (15-min slots, EUR/MWh) and stores
the raw market price as feed_in_tariff_wh (EUR/Wh) — no import charges/VAT.
    
Slots beyond the day-ahead horizon are left to the consumer's forward-fill
(FeedInTariffImport behaviour).
    
5 unit tests + opt-in live smoke (EOS_DVHUB_ONLINE_LIVE=1, passing).

Waits for the following PRs to be merged:

- #1191


NormannK:
Looking at the current data flow:

```text
FeedInTariffDvhubOnline._update_data()
    → stores only the values returned by DVhub
    → GeneticOptimizationParameters.prepare()
    → prediction.key_to_array(..., fill_method="ffill")
    → DataSequence.key_to_array()
    → resampled.ffill()
```

When the requested 48-hour prediction horizon extends beyond the available DVhub data, the missing portion of that horizon is therefore forward-filled with the last available hourly resampled tariff value.

Since DVhub provides 15-minute prices, `key_to_array()` first averages the available values into hourly buckets. It then repeats the final hourly bucket across all remaining missing hours. This could materially distort optimization decisions, especially when the final available hour has an unusually high or negative price.

Should this provider explicitly extend the series using a defined fallback strategy, or fail when the available data does not cover the requested horizon?

thanks chatgpt to get my thoughts into a coherent sentance.

src/akkudoktoreos/prediction/feedintariffdvhubonline.py
```diff
@@ -0,0 +1,147 @@
...
+        try:
+            dvhub_data = self._request_forecast(
+                start_date=start_date, end_date=end_date, force_update=force_update
+            )  # type: ignore[call-arg]
+            series_data = self._parse_data(dvhub_data)
+            if series_data.empty:
+                raise ValueError("No dvhub.online feed-in tariff data available")
+            self.highest_orig_datetime = series_data.index.max()
+            await self.key_from_series("feed_in_tariff_wh", series_data)
+        except Exception as exc:
```
NormannK, line 136: Could we avoid catching every Exception here?

Once highest_orig_datetime has been set, this also suppresses permanent or unexpected failures, such as HTTP 4xx responses, invalid JSON/API schema changes, parsing or storage errors, and programming defects. In those cases, the provider would silently retain stale prices, which may then be forward-filled and treated as valid tariff data.

b0661: This is the original code from @chloepriceless. I opened issue #1204 to track this deficiency. Should be merged with bug tracked.

b0661:
> Looking at the current data flow:
> 
> ```
> FeedInTariffDvhubOnline._update_data()
>     → stores only the values returned by DVhub
>     → GeneticOptimizationParameters.prepare()
>     → prediction.key_to_array(..., fill_method="ffill")
>     → DataSequence.key_to_array()
>     → resampled.ffill()
> ```
> 
> When the requested 48-hour prediction horizon extends beyond the available DVhub data, the missing portion of that horizon is therefore forward-filled with the last available hourly resampled tariff value.
> 
> Since DVhub provides 15-minute prices, `key_to_array()` first averages the available values into hourly buckets. It then repeats the final hourly bucket across all remaining missing hours. This could materially distort optimization decisions, especially when the final available hour has an unusually high or negative price.
> 
> Should this provider explicitly extend the series using a defined fallback strategy, or fail when the available data does not cover the requested horizon?
> 
> thanks chatgpt to get my thoughts into a coherent sentance.

This is the original code from @chloepriceless. I opened issue https://github.com/Akkudoktor-EOS/EOS/issues/1204 to track this deficiency. Should be merged with bug tracked.

PR #1194, opened 2026-07-26
Fix battery/EV residual value and per-step degradation cost

jsb1:
Hello team, Moin Andreas,

I’m currently working on implementing an alternative optimizer based on “Dynamic Programming.” It’s already working quite well and is extremely fast, but I don’t want to submit it as a pull request just yet, since the structural changes are quite significant. The code is also far from clean right now.
For those interested: [https://github.com/jsb1/EOS/tree/feat/dp-solver]

During the DP implementation, I noticed a few things:

1. At the end of the simulation horizon, the value of additional or missing battery charge (including EVs) is not taken into account. Since I can’t determine this value precisely, I’ve used the average between the possible feed-in tariff and the electricity price at the end of the time window as an estimate. It’s simple, but not entirely clean. If the electricity price is negative at that time and beyond, the value is essentially zero. However, we can’t know that at the start of the simulation.

2. The potential battery degradation should be considered for each charging operation within the simulated period, not only at the end. Of course, charge and discharge activities within a single hour are still not captured. But this approach might be a bit better.

Best regards,

Jörg


b0661:
> I’m currently working on implementing an alternative optimizer based on “Dynamic Programming.” It’s already working quite > well and is extremely fast, but I don’t want to submit it as a pull request just yet, since the structural changes are quite significant. The code is also far from clean right now.
> For those interested: [https://github.com/jsb1/EOS/tree/feat/dp-solver]

Great news. Just look at #1190 on how to add another optimization algorithm to EOS.

b0661:
> During the DP implementation, I noticed a few things:

>    At the end of the simulation horizon, the value of additional or missing battery charge (including EVs) is not taken into account. Since I can’t determine this value precisely, I’ve used the average between the possible feed-in tariff and the electricity price at the end of the time window as an estimate. It’s simple, but not entirely clean. If the electricity price is negative at that time and beyond, the value is essentially zero. However, we can’t know that at the start of the simulation.

 >   The potential battery degradation should be considered for each charging operation within the simulated period, not only at the end. Of course, charge and discharge activities within a single hour are still not captured. But this approach might be a bit better.

As we now somehow retiring the current genetic algorithm. Is this something you want to see in the new genetic algorithm as @drbacke is developing in https://github.com/Akkudoktor-EOS/EOS/tree/feat/direct-marketing-battery-grid-export?

jsb1:
That's fine, I also abandoned my DP approach and try do-mpc for now.

I would like to see some refactoring on the simulation engine to be able to plug in alternative solvers. 

I can do in if You want.

NormannK:
fyi #1190 was merged

PR #1195, opened 2026-07-28
chore: adapt deprecated endpoints to 15-minutes predictions

b0661:
Ensure the deprecated endpoints get /strompreis, post /gesamtlast, get /gesamtlast_simple,
get /pvforecast to work on 1-hour intervalls even if the prediction provides 15-minutes
intervall data. This keeps the interface compliant to the legacy functionality.
    
Besides this adaptations several other improvements and fixes are included in this PR.
    
* feat: extend data management method key_to array by resample_method
    
  One can now define the resample method on how to aggregate the values in an interval
  for resampling. Three methods are provided:
    
  - "first": Use the first value in each interval.
  - "mean": Compute the arithmetic mean of all samples in each interval.
  - "interval_mean": Compute the time-weighted mean assuming each value remains valid
     until the next timestamp (piecewise-constant signal).
    
* feat: extend the get /v1/prediction/dataframe endpoint with resampling parameters
    
  Make all parameters for resampling available at the endpoint.
    
* feat: extend the get /v1/prediction/list endpoint with resampling parameters
    
  Make all parameters for resampling available at the endpoint.
    
* feat: add new delete /v1/prediction/range endpoint
    
  The endpoint allows to delete prediction values for a given time span.
    
* fix: adapt for PVForecastAkkudoktor server side cache handling
    
  /api.akkudoktor/forecast does it's own caching on requests. Call it with slightly
  randomized request values to avoid getting cached values in the case we need fresh
  data. The requests are anyway rate limited to one request per hour on our side.
    
* chore: add core.types
    
  This module centralizes reusable type definitions shared across multiple
  packages. Defining common types here avoids duplication of complex type
  annotations (such as Literal aliases), ensures consistent typing across the
  code base, and helps prevent circular import dependencies between modules.
    
* chore: extend cache testing
    
* chore: add system test for deprecated /strompreis endpoint
    
* chore: add unit test module for server endpoints
    
  Add a new test module to do unit tests on server endpoints. First test added
  for deprecated get /strompreis endpoint.

Waits for the following PRs to be merged:

- #1191
- #1193


src/akkudoktoreos/server/eos.py
```diff
@@ -973,16 +978,77 @@ async def fastapi_prediction_dataframe_get(
...
+            Defaults to end datetime of latest prediction.
+        interval (Optional[str]): Time duration for each interval.
+            Defaults to 1 hour.
+        fill_method (str): Method to handle missing values during resampling.
+
+            - 'linear': Linearly interpolate missing values (for numeric data only).
+            - 'time': Interpolate missing values (for numeric data only).
+            - 'ffill': Forward fill missing values.
+            - 'bfill': Backward fill missing values.
+            - 'none': Defaults to 'linear' for numeric values, otherwise 'ffill'.
```
NormannK: `"none"` as a supported value, but the `FillMethod` type in `src/akkudoktoreos/core/types.py` only allows:

```python
"linear", "time", "ffill", "bfill"
```

Because this type is used by the FastAPI query parameter, `fill_method=none` fails validation with HTTP 422 before reaching `key_to_array()`, even though `key_to_array()` supports `"none"`.

Please add `"none"` to `FillMethod`

NormannK (approved):

src/akkudoktoreos/server/eos.py
b0661: The string "none" is not supported. Changed the description.

PR #1205, opened 2026-07-29
chore: adapt pdf visualization

### LINKED ISSUE #1206: [BUG]: data align problem
### Describe the issue:

src/akkudoktoreos/core/dataabc.py

`        reference_index = pd.date_range(
            start=start_datetime,
            end=end_datetime,
            freq=interval,
            inclusive="left",
        )
`
~LIne 2615

When `align_to_interval=True`, `key_to_array()` aligns its buckets to epoch boundaries, but this dataframe index still starts at the original `start_datetime`.

For example, with `start_datetime=10:07` and `interval=15 minutes`, the returned array is based on buckets at `10:15, 10:30, ...` after truncation, while this index labels those values as `10:07, 10:22, ...`. Since only the array length is checked, the mismatch is not detected and the dataframe contains incorrectly timestamped values.

Fix: 
construct `reference_index` using the same alignment and truncation rules as `key_to_array()`. 

```python

```

### Error message:

```shell

```

### Version information:

meh

### LINKED ISSUE #1210: [BUG]: Concurrent PDF requests can corrupt or overwrite visualization output
### Description

The `/visualization_results.pdf` endpoint generates the visualization PDF on demand using a fixed output path:

```python
visualize_file = output_path / "genetic0_solution_visualisation.pdf"
```

Every request writes to this same file before returning it through `FileResponse`.

There is currently no synchronization, request-specific temporary file, or atomic file replacement. As a result, concurrent requests can write to and read from the same PDF file simultaneously.

### Affected code

File:

```text
src/akkudoktoreos/server/eos.py
```

Relevant endpoint:

```python
@app.get("/visualization_results.pdf", response_class=PdfResponse, tags=["optimize"])
def get_pdf() -> PdfResponse:
    genetic0_solution = get_ems().genetic0_solution()
    if genetic0_solution is None:
        raise HTTPException(
            status_code=400,
            detail="Optimize error: no solution stored by run.",
        )

    output_path = get_config().general.data_output_path
    if output_path is None or not output_path.is_dir():
        raise HTTPException(
            status_code=404,
            detail=f"Output path does not exist: {output_path}.",
        )

    visualize_file = output_path / "genetic0_solution_visualisation.pdf"

    genetic0_prepare_visualize(
        solution=genetic0_solution,
        filename=str(visualize_file),
    )

    return PdfResponse(visualize_file)
```

Related implementation:

```text
src/akkudoktoreos/optimization/genetic0/genetic0visualize.py
```

### Problem

When two or more clients request `/visualization_results.pdf` concurrently:

1. Both requests generate a PDF at the same filesystem path.
2. One request may truncate or overwrite the file while another request is still generating it.
3. `FileResponse` may begin streaming the file while another request modifies it.
4. The returned PDF may be incomplete, corrupted, inconsistent, or based on a different optimization solution.
5. Matplotlib and `PdfPages` resources may also interact through shared files and global plotting state.

### Steps to reproduce

1. Run an optimization so that a `GENETIC0` solution is available.
2. Send several concurrent requests to the endpoint.

Example:

```python
from concurrent.futures import ThreadPoolExecutor

import requests

URL = "http://localhost:8503/visualization_results.pdf"


def download_pdf(index: int) -> tuple[int, int, bytes]:
    response = requests.get(URL, timeout=60)
    return index, response.status_code, response.content


with ThreadPoolExecutor(max_workers=10) as executor:
    results = list(executor.map(download_pdf, range(20)))

for index, status, content in results:
    print(index, status, len(content), content[:8])
```

3. Compare the returned file sizes and validate each response with a PDF parser.
4. Repeat the test while a new optimization solution is being stored.

### Actual behavior

Multiple requests write to the same `genetic0_solution_visualisation.pdf` file.

Depending on timing, responses may:

* contain truncated or malformed PDF data;
* have different or unexpectedly small file sizes;
* fail PDF validation;
* contain output belonging to another request or optimization solution;
* fail with filesystem or Matplotlib-related errors.

The issue may be intermittent because it depends on request timing.

### Expected behavior

Each request should receive a complete PDF generated from one consistent optimization solution.

Concurrent requests must not overwrite, truncate, or modify files used by other requests.

### Impact

This can affect any deployment where multiple users, browser tabs, dashboards, reverse proxies, or automated clients request the visualization endpoint at the same time.

Potential consequences include:

* corrupted downloads;
* inconsistent visualization data;
* intermittent server errors;
* disclosure of output generated for a different request;
* orphaned or locked files;
* difficult-to-reproduce production failures.

### Suggested solution

Generate the PDF into a unique temporary file for each request.

Possible implementation:

```python
from pathlib import Path
from tempfile import NamedTemporaryFile

from fastapi.background import BackgroundTask


def remove_file(path: Path) -> None:
    path.unlink(missing_ok=True)


@app.get("/visualization_results.pdf", response_class=PdfResponse, tags=["optimize"])
def get_pdf() -> PdfResponse:
    solution = get_ems().genetic0_solution()
    if solution is None:
        raise HTTPException(
            status_code=400,
            detail="Optimize error: no solution stored by run.",
        )

    output_path = get_config().general.data_output_path
    if output_path is None or not output_path.is_dir():
        raise HTTPException(
            status_code=404,
            detail=f"Output path does not exist: {output_path}.",
        )

    with NamedTemporaryFile(
        prefix="genetic0_visualisation_",
        suffix=".pdf",
        dir=output_path,
        delete=False,
    ) as temporary_file:
        pdf_path = Path(temporary_file.name)

    try:
        genetic0_prepare_visualize(
            solution=solution,
            filename=str(pdf_path),
        )
    except Exception:
        pdf_path.unlink(missing_ok=True)
        raise

    if not pdf_path.is_file():
        raise HTTPException(
            status_code=500,
            detail="Visualization PDF could not be generated.",
        )

    return PdfResponse(
        pdf_path,
        filename="visualization_results.pdf",
        background=BackgroundTask(remove_file, pdf_path),
    )
```

If generated PDFs should be cached, generation should instead use:

* a lock around cache creation;
* a unique temporary file;
* an atomic `Path.replace()` after successful generation;
* a cache key tied to the specific solution.

### Additional considerations

* Ensure every `PdfPages` instance is explicitly closed.
* Avoid creating an initial `PdfPages` object that is later replaced without being closed.
* Avoid fixed SVG filenames when multiple requests can generate charts concurrently.
* Consider protecting Matplotlib rendering with a lock if the selected Matplotlib operations are not thread-safe in the server environment.
* Snapshot the optimization solution before generation so that one report cannot mix data from different optimization runs.

### Acceptance criteria

* Concurrent requests produce valid and complete PDFs.
* Each request uses an isolated output file or a safely generated immutable cache file.
* Temporary files are removed after the response completes.
* Failed generation does not leave partial PDFs behind.
* A concurrency test sends multiple simultaneous requests and validates every returned PDF.
* Generation remains correct while a new optimization solution is being stored.

```python

```

### Error message:

```shell

```

### Version information:

for https://github.com/Akkudoktor-EOS/EOS/pull/1205

b0661:
Change PDF visualization to be created on demand and per optimization algorithm. The PDF
for the GENETIC0 optimization is provided by the /visualization_results.pdf endpoint.
There is no change in the interface.

By this the optimization algorithm is offloaded from the PDF generation which spares some
time.

To cope with several users may call the /visualization_results.pdf endpoint at the same
time the PDF is generated on the fly without any intermediate file taking the stored
GENETIC0 solution as an input. SVG picture generation is removed as this would again
create intermediate files. Chart pictures can easily be taken from the PDF.

To allow on demand creation of the optimization results visualization the optimisation
solution stored is extended by several new attributes. To keep the deprecated
/optimize endpoint compatible the optimization solution is stripped to the legacy
content before returned. Due to the extension of the solution the optimization tests were
adapted to cover the extended content.

The optimization tests are adapted to test the generated visualization report by
the pypdf reader. Pypdf is added to the development dependencies.

Besides the adaptation several fixes and improvements are added:

* feat: extend /v1/prediction/series endpoint by resampling and filling

  Add parameters for resampling and filling. Add the processing parameter
  to control wether raw data or resampled data shall be returned.

* feat: extend /v1/measurement/series endpoint by resampling and filling

  Add parameters for resampling and filling: Add the processing parameter
  to control wether raw data or resampled data shall be returned.

* feat: standardize and improve API error response

  Use FASTApi exception handlers to provide a standardized API exception handling.
  All exceptions are logged.

  Exception traces are only returned if the new logging configuration parameter
  logging.api_logging_level is set to "DEBUG" or "TRACE". Avoids unwanted leackage
  of server internals on exceptions.

* fix: align to intervall when resampling

  Ensure resampling is aligned to interval also when the buckets are shifted due to the
  align_to_intervall parameter is set.

* chore: make dropna mandatory and default to True

* chore: refactor key_to_xxx data management methods

  Make key_to_series the central method for data resampling and fill.
  Add a new key_to_raw_series to retrieve the data as it is stored
  (without resampling and filling).

  Users of key_to_series were mostly moved to key_to_raw_series as this resembles
  the former interface. Especially in predictions and tests this was done.

* chore: create test data sub-directory for each optimization algorithm

  To prevent cluttering the test data directory and ease test data management for
  optimization algorithms each algorithm got it's own sub-directory. The current
  test data was moved to these sub-directories.

* chore: update version

Closes: #1206
Closes: #1210 

This pull request waits for:

- #1193
- #1195


tests/test_system.py
```diff
@@ -581,7 +639,10 @@ def test_measurement_high_frequency_and_duplicates(self, server_setup_for_class)
         # ----------------------------------------------------------------------
         result = requests.get(
             f"{server}/v1/measurement/series",
-            params={"key": "pv1_emr_kwh"},
+            params={
+                "key": "pv1_emr_kwh",
+                "processimg": "raw",
```
NormannK: typo. processiMg

src/akkudoktoreos/server/eos.py
```diff
@@ -735,25 +747,149 @@ def fastapi_measurement_keys_get() -> list[str]:
...
+        Query(
+            description="Snap resample origin to the nearest UTC epoch-aligned boundary of interval."
+        ),
+    ] = False,
+    processing: Annotated[
+        SeriesProcessing,
+        Query(
+            description="Processing mode. 'raw' returns original measurement data without resampling or filling."
+        ),
+    ] = SeriesProcessing.RESAMPLED,
```
NormannK: Could we keep `RAW` as the default for backward compatibility?

Before this change, calling `/v1/measurement/series` without additional parameters returned the stored timestamps and values unchanged. With `RESAMPLED` as the default, the same existing request now performs hourly resampling using `mean` and may fill missing values through interpolation. This silently changes both timestamps and values for existing API clients while still returning a successful response.

I suggest defaulting to `SeriesProcessing.RAW` and requiring callers to explicitly request `processing=resampled`. The same applies to `/v1/prediction/series`.

src/akkudoktoreos/server/eos.py
```diff
@@ -735,25 +747,149 @@ def fastapi_measurement_keys_get() -> list[str]:
...
-        # Re-raise HTTP exceptions
-        raise
     except Exception as e:
         # Log unexpected errors
         trace = "".join(traceback.TracebackException.from_exception(e).format())
-        logger.exception(f"Unexpected error retieving measurement: {key}")
+        logger.exception(f"Unexpected error retrieving measurement: {key}")
         raise HTTPException(
             status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
             detail=f"Internal server error:\n{e}\n{trace}",
```
NormannK: Errors caused by query parameters, such as an invalid `interval`, may raise `ValueError` or a similar validation exception inside `key_to_series()`. The broad `except Exception` converts those client-input errors into HTTP 500 responses. In addition, including `trace` in `detail` exposes internal file paths, implementation details, and stack frames through the public API.

```python
except (TypeError, ValueError) as exc:
    raise HTTPException(status_code=422, detail=str(exc)) from exc
except HTTPException:
    raise
except Exception as exc:
    logger.exception("Unexpected error retrieving measurement: %s", key)
    raise HTTPException(
        status_code=500,
        detail="Internal server error.",
    ) from exc
```

The same issue occurs in `fastapi_prediction_series_get`.

src/akkudoktoreos/server/eos.py
```diff
@@ -957,9 +1184,36 @@ async def fastapi_prediction_series_get(
...
+                boundary=boundary,
+                align_to_interval=align_to_interval,
+            )
+    except Exception as e:
+        # Log unexpected errors
+        trace = "".join(traceback.TracebackException.from_exception(e).format())
+        logger.exception(f"Unexpected error retrieving prediction: {key}")
+        raise HTTPException(
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail=f"Internal server error:\n{e}\n{trace}",
```
NormannK: same as R888

tests/test_system.py
b0661: Fixed

src/akkudoktoreos/server/eos.py
b0661: Default set to raw.

src/akkudoktoreos/server/eos.py
b0661: Query parameters are now explicitly checked. New standardize API error handling is applied.

src/akkudoktoreos/server/eos.py
b0661: Query parameters are now explicitly checked. New standardize API error handling is applied.

src/akkudoktoreos/server/rest/error.py
```diff
@@ -1,4 +1,109 @@
...
+    if cause is not None:
+        logger.exception(cause)
+
+        if get_config().logging.api_level in ("TRACE", "DEBUG"):
+            body["traceback"] = traceback.format_exception(cause)
+
+    return JSONResponse(
+        status_code=status,
+        media_type="application/problem+json",
+        content=body,
```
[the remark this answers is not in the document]
b0661: Exposure is guarded by logging.api_logging_level

PR #1208, opened 2026-07-30
fix(eosdash): render prediction/plan datetimes in correct local time

cornim:
## Problem
EOSdash prediction and plan charts show all datetimes shifted into the future by the local UTC offset (e.g. +2h for Europe/Berlin in summer).

## Root cause
The dashboards strip the timezone from the UTC-aware `date_time` values, handing Bokeh naive *local* timestamps. Bokeh (`x_axis_type="datetime"`) serializes naive datetimes as epoch-ms interpreted as UTC and renders axis ticks in the *browser's* local timezone, applying the offset a second time — a double shift.

## Fix
Feed Bokeh the true UTC instant (naive-UTC). Bokeh renders it in the browser's local timezone once, so a browser in the EOS-configured timezone shows the correct local wall-clock. Bokeh 3.9.1 exposes no server-side axis timezone, so this is the deterministic fix for the common case where the viewing browser and the EOS timezone match.

- `prediction.py`: parse as UTC, drop tz (keep UTC clock); use the config timezone for the axis label.
- `plan.py`: stop converting to local before dropping tz.


src/akkudoktoreos/server/dash/prediction.py
```diff
@@ -245,9 +245,13 @@ def Prediction(eos_host: str, eos_port: Union[str, int], data: Optional[dict] =
     except Exception as err:
         return Error(f"Can not retrieve predictions from {server}: {err}")
 
-    # Remove time offset from UTC to get naive local time and make bokeh plot in local time
-    date_time_tz = predictions["date_time"].dt.tz
-    predictions["date_time"] = pd.to_datetime(predictions["date_time"]).dt.tz_localize(None)
+    # Feed Bokeh the true UTC instant (naive-UTC). Bokeh renders datetime axes in the
+    # viewing browser's local timezone, so a browser in the EOS config timezone shows
+    # correct local wall-clock. (Bokeh 3.9.1 has no server-side axis timezone.)
+    date_time_tz = config["general"]["timezone"]
```
NormannK: its minor but `general.timezone` is a optional value and can be `None` when either coordinate is missing. It also currently returns `None` for valid coordinates where latitude or longitude is `0`, because the computed property uses a truthiness check. Could we provide a fallback here, or fix the computed property to check `is not None`, so the axis label cannot become `Datetime [localtime None]`?

NormannK (approved):

src/akkudoktoreos/server/dash/prediction.py
NormannK: `date_time_tz = config["general"].get("timezone") or "Europe/Berlin"`
would do.

src/akkudoktoreos/server/dash/prediction.py
```diff
@@ -245,9 +245,13 @@ def Prediction(eos_host: str, eos_port: Union[str, int], data: Optional[dict] =
...
+    predictions["date_time"] = pd.to_datetime(predictions["date_time"], utc=True).dt.tz_localize(
```
b0661, line 258: The datetime shall be that of the local time at the location of the system (not the local time of the browser), This is also displayed on the time axis.

src/akkudoktoreos/server/dash/plan.py
```diff
@@ -172,10 +172,10 @@ def SolutionCard(solution: OptimizationSolution, config: SettingsEOS, data: Opti
...
     else:
         date_time_tz = config.general.timezone
-    # Ensure original date_time is parsed as UTC and convert to local time
-    df["date_time_local"] = (
-        pd.to_datetime(df["date_time"], utc=True).dt.tz_convert(date_time_tz).dt.tz_localize(None)
-    )
+    # Feed Bokeh the true UTC instant (naive-UTC). Bokeh renders datetime axes in the
+    # viewing browser's local timezone, so a browser in the EOS config timezone shows
+    # correct local wall-clock. (Bokeh 3.9.1 has no server-side axis timezone.)
+    df["date_time_local"] = pd.to_datetime(df["date_time"], utc=True).dt.tz_localize(None)
```
b0661, line 184: The datetime shall be that of the local time at the location of the system (not the local time of the browser), This is also displayed on the time axis.

src/akkudoktoreos/server/dash/prediction.py
cornim: Good catch, applied as suggested (also fixed the underlying computed-field truthiness bug in `GeneralSettings.timezone`, which returned `None` for lat/long == 0.0 as well). Pushed in 4d33a1c.

src/akkudoktoreos/server/dash/prediction.py
cornim: Agreed that ideally the axis should show the EOS system's configured local time regardless of the viewer's browser timezone, not just "the common case where browser and EOS timezone match" as I put it in the PR description. However, achieving that robustly needs a client-side fix (e.g. a Bokeh `CustomJSTickFormatter` using `Intl.DateTimeFormat` with an explicit IANA `timeZone`, since BokehJS's default datetime axis always renders using the browser's local zone) rather than anything we can do purely server-side with pandas/tz_convert - and that's what caused the original double-shift bug this PR fixes.

I'd suggest landing this PR as-is to fix the double-shift regression (verified fix, plus two related fixes: the `None`-timezone fallback and a computed-field truthiness bug for lat/long == 0.0), and opening a separate follow-up issue for the browser-vs-system-timezone concern so it can be scoped and reviewed on its own. Let me know if that sounds reasonable and I'll file the ticket.

src/akkudoktoreos/server/dash/plan.py
cornim: see above

src/akkudoktoreos/server/dash/prediction.py
NormannK: yeah lat/long == 0.0  i saw too bat than again who is living on the aquator.... :-D

PR #1209, opened 2026-07-30
fix(database): open backend so records actually persist and reload

cornim:
## Problem
Measurement and prediction records are never persisted to the configured database backend (e.g. LMDB). They only survive while the process runs; every restart loses the accumulated history. For `LoadAkkudoktorAdjusted` this silently zeroes the measurement-based adjustment (the adjusted load forecast collapses onto the raw mean).

## Root cause
`db_enabled` is defined as `database.is_open`, and every database code path (init, load, save, insert) is guarded behind `db_enabled`. Nothing ever opens the backend first — the only lazy open (via `_run_db`) is unreachable because those calls sit behind the same guard. The backend stays closed, `db_enabled` stays False, and all writes fall through to the JSON file fallback, which only holds the current in-memory snapshot and is not reloaded into the record store on startup.

## Fix
Explicitly open the configured backend in `_db_ensure_initialized` before the `db_enabled` gate, wrapped in try/except so a failure degrades gracefully to file storage.

## Verification
A/B test (identical persisted config, push -> save -> restart): without the fix the backend reports `enabled=false` and data is lost on restart; with the fix the backend is enabled, records persist to LMDB, and reload correctly after restart.

### DIFF
--- a/src/akkudoktoreos/core/databaseabc.py
+++ b/src/akkudoktoreos/core/databaseabc.py
@@ -635,8 +635,8 @@ class DatabaseRecordProtocolMixin(
         #
         # Skip disabled providers (None/"NoDB"): persistence is intentionally off and
         # `NoDB.is_open` is always False, so opening it would run on every call. Only
-        # attempt the open once; on failure we log and fall back to file storage
-        # without retrying (and re-logging) on every subsequent record operation.
+        # attempt the open once; on failure we log without retrying (and re-logging)
+        # on every subsequent record operation.
         provider = self.config.database.provider
         if (
             provider not in (None, "NoDB")
@@ -648,8 +648,7 @@ class DatabaseRecordProtocolMixin(
                 await self.database.open(namespace=self.db_namespace())
             except Exception:
                 logger.exception(
-                    f"Could not open database backend for namespace '{self.db_namespace()}'; "
-                    "falling back to file storage."
+                    f"Could not open database backend for namespace '{self.db_namespace()}'."
                 )
 
         if not self._db_storage_initialized and self.db_enabled:
--- a/src/akkudoktoreos/core/databaseabc.py
+++ b/src/akkudoktoreos/core/databaseabc.py
@@ -620,6 +620,11 @@ class DatabaseRecordProtocolMixin(
             self._db_metadata = None
             self._db_storage_initialized: bool = False
 
+            # Whether an explicit backend open has already been attempted.
+            # One-shot per init so a closed/failed backend is not retried on
+            # every record operation. Re-armed when DB state is reset.
+            self._db_open_attempted: bool = False
+
             self._db_initialized: bool = True
 
         # Ensure the configured database backend is actually opened.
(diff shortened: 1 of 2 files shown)


NormannK:
This retries `database.open()` on every call to `_db_ensure_initialized()` whenever the backend remains closed.

That includes two important cases:

1. `NoDB.is_open` always returns `False`, so configurations with `provider=None` or `provider="NoDB"` repeatedly enter this path even though persistence is intentionally disabled.
2. If LMDB or SQLite fails to open, the exception is caught but no failure state is recorded. Every subsequent insert, load, or save operation retries the open and logs another exception.

Please skip opening for disabled providers and prevent repeated immediate retries after an opening failure, for example with an initialization-attempt flag or retry backoff.

At minimum, the provider check could look like:

```python
provider = self.config.database.provider

if provider not in (None, "NoDB") and not self.database.is_open:
    ...
```

Please also add coverage confirming that `NoDB` does not repeatedly call `open()` and that an unavailable backend does not produce an exception on every record operation.

cornim:
Thanks for the review @NormannK. Addressed both points in 68cecf3:

- **Disabled providers**: `_db_ensure_initialized` now reads `self.config.database.provider` and skips the open entirely when it is `None`/`"NoDB"`, so persistence-disabled configs no longer call `open()` on every operation.
- **Failed open retries**: added a one-shot `_db_open_attempted` flag (re-armed only on DB state reset), so a closed/unavailable backend is opened at most once — the exception is caught, logged once, and subsequent inserts/loads/saves fall through to file storage without retrying or re-logging.

Added coverage in `tests/test_dataabcsequencedb.py`: one test asserts `NoDB` never calls `open()`, another asserts a failing backend does not raise per record operation and is opened at most once (`call_count == 1`).

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -620,8 +620,38 @@ async def _db_ensure_initialized(self) -> None:
...
+            and not self.database.is_open
+            and not self._db_open_attempted
+        ):
+            self._db_open_attempted = True
+            try:
+                await self.database.open(namespace=self.db_namespace())
+            except Exception:
+                logger.exception(
+                    f"Could not open database backend for namespace '{self.db_namespace()}'; "
+                    "falling back to file storage."
```
b0661: File storage fall back is not a general feature of the database implementation. It is a special "legacy" case for measurements and the load and save operations. Just skip this part of the exception message.

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -620,8 +620,38 @@ async def _db_ensure_initialized(self) -> None:
...
 
+        # Ensure the configured database backend is actually opened.
+        # `db_enabled` reflects `database.is_open`, and every DB code path is guarded
+        # behind `db_enabled`. Without an explicit open here the backend is never
+        # opened (lazy open via `_run_db` is unreachable), so records would only ever
+        # persist to the JSON file fallback and never reload into memory on startup.
+        #
+        # Skip disabled providers (None/"NoDB"): persistence is intentionally off and
+        # `NoDB.is_open` is always False, so opening it would run on every call. Only
+        # attempt the open once; on failure we log and fall back to file storage
```
b0661: File storage fall back is not a general feature of the database implementation. It is a special "legacy" case for measurements and the load and save operations. Just skip this part of the comment.

b0661 (changes requested):

src/akkudoktoreos/core/databaseabc.py
cornim: Done in 0c73cf9 — removed the "falling back to file storage." line from the open-failure log message.

src/akkudoktoreos/core/databaseabc.py
cornim: Done in 0c73cf9 — dropped the file-storage-fallback wording from the comment.

b0661 (approved):

PR #1211, opened 2026-07-31
fix(dataabc): drop NaN values in key_to_dict/key_to_lists

cornim:
## Summary

`key_to_dict` and `key_to_lists` filtered out NaN values with
`getattr(record, key, None) == float("nan")`. Since `NaN != NaN`, this
comparison is always `False`, so NaN values were **never** dropped when
`dropna=True` (the default). They leaked into `key_to_series` / `key_to_array`
and downstream resampling/interpolation, silently corrupting results.

## Fix

Use `pd.isna(...)` to detect NaN, matching the convention already used
elsewhere in the module.

## Tests

Added red/green regression tests (`test_key_to_dict_dropna_removes_nan`,
`test_key_to_lists_dropna_removes_nan`) that verify NaN records are dropped
with the default `dropna=True` and kept with `dropna=False`. Both fail against
the unfixed code and pass with the fix.

## Checks

- `pre-commit run --files ...` — all hooks pass (isort, ruff, ruff format, mypy)
- affected pytest suites green


b0661:
Good catch. Thanks for the PR and the test!!!.

NormannK:
ahh nice. i saw that too and was not sure. now I am :-) will be merged after the #1205 and #1209 
this will need a rebase after #1205

NormannK (approved):

PR #1212, opened 2026-07-31
feat(elecprice): model charges/fees as ordered component list

### LINKED ISSUE #392: [ENH]: Charges and Fees correct
### Link to discussion and related issues

```python
Fees are percentual, at the moment there is only one fix charge added, which is not correct. We should implement the single parts of fees ...
```

### Proposed implementation

```python

```

cornim:
> **Note:** This PR is meant as a **basis for discussion** and can of course be modified — naming, structure, defaults, and scope are all open to change based on your feedback.

## Summary

Closes #392.

The German electricity price is not just "spot price + one fixed charge". It is a build-up of several components — grid fees (Netzentgelt/Messstellenbetrieb), concession fee (Konzessionsabgabe), electricity tax (Stromsteuer), various surcharges (Offshore/EnWG …), and finally 19 % VAT applied on top of everything.

Today EOS only supports a single fixed `charges_kwh` plus a `vat_rate` multiplier, and the VAT handling is inconsistent between providers:

- `ElecPriceAkkudoktor` adds `charges_kwh` but **ignores** `vat_rate` entirely.
- `ElecPriceEnergyCharts` applies `(market + charges) * vat_rate`, but **only when `charges_kwh > 0`** (so VAT is silently dropped for spot-only setups).
- `Tibber`, `Fixed`, `Import` ignore both.

## What this changes

Replaces `charges_kwh` + `vat_rate` with an **ordered list of charge components** (`elecprice.charges`). Each component is either:

- **`fixed`** — an absolute amount per kWh, or
- **`percent`** — a percentage add-on computed on a configurable **basis** (referenced preceding components and/or the literal `"market"` for the spot price). If `basis` is omitted, the percentage applies to the full accumulated price so far (market price plus all preceding add-ons).

Components are processed **first → last**, so the German build-up can be expressed directly:

```json
{
  "elecprice": {
    "charges": [
      {"name": "Netzentgelt",       "type": "fixed",   "amount": 0.1153},
      {"name": "Konzessionsabgabe", "type": "fixed",   "amount": 0.018},
      {"name": "Stromsteuer",       "type": "fixed",   "amount": 0.0205},
      {"name": "Umlagen",           "type": "fixed",   "amount": 0.0158},
      {"name": "MwSt",              "type": "percent", "amount": 0.19}
    ]
  }
}
```

Time-variable surcharges are intentionally out of scope for now — all components are assumed constant across the day.

**Backward-compatible behavior:** the previous functionality can be fully reproduced with this scheme. A single fixed charge with VAT applied on top — the old `(market + charges_kwh) * vat_rate` — is simply a `fixed` component followed by a `percent` component (e.g. `[{"type": "fixed", "amount": 0.21}, {"type": "percent", "amount": 0.19}]`). This is exactly what the migration below produces, so nobody loses their current setup.

**Negative prices:** based on my research into how German dynamic tariffs work, charges are added regardless of the sign of the spot price (no clamping) — fixed components always apply, a negative spot price simply lowers the hour's price, and the end price only turns negative if the negative spot exceeds the sum of the fixed parts. A `percent` (VAT) component multiplies the running total, so it scales the net amount including any negative part. The current implementation already behaves this way.

## Design decisions (open for discussion)

- **Centralized**: the calculation lives in `ElecPriceCommonSettings.apply_charges` and is exposed via the `ElecPriceProvider` base class, so every provider behaves identically (also fixes the inconsistent VAT handling above).
- **Applied by** Akkudoktor, EnergyCharts, Fixed and Tibber. **`ElecPriceImport` is intentionally left untouched** (imported data is assumed final). If a provider already returns a full consumer price (e.g. Tibber), users simply leave `charges` unset and it's a no-op.
- **Percent basis default**: omitting `basis` means "everything so far", which matches how VAT works in Germany.

## Migration

Legacy `charges_kwh`/`vat_rate` configs are migrated automatically into an equivalent `charges` list — a `fixed` component followed by a `percent` component of `vat_rate - 1` (so `0.21` + `1.19` → `[{fixed 0.21}, {percent 0.19}]`). The old fields are removed. Because migration keys off the auto-computed version hash, existing configs migrate on upgrade without any manual version bump.

## Tests

- New `tests/test_elecpricecharges.py` — unit tests for `apply_charges`: fixed-only, percent with default/explicit basis, `"market"` basis, multiple percents, ordering, empty/None no-op, unknown-basis error, and the full German build-up from the issue.
- New migration tests (`test_configmigrate.py`) covering the `charges_kwh`/`vat_rate` → `charges` combination (incl. default VAT and the `None` no-op case).
- New end-to-end provider test (`test_elecpriceakkudoktor.py`) proving charges are applied on top of the fetched market price.
- Updated fixtures, `openapi.json`, generated config docs, and `docs/akkudoktoreos/prediction.md`.

## Checks

- `pre-commit run --files …` (isort, ruff, ruff format, mypy, …) passes on all changed files.
- Targeted `pytest` suites pass (elecprice providers, charges, migration, config, prediction, feedintariff).


b0661:
Good point.  

Maybe we can even make this more flexible for other schemes of electricity fees. Other countries have other schemes how to calculate the electricity fees. These schemes may even need complex calculations for e.g. peek load forecast or dynamic grid fee calculations.

My idea is that we have a new class of prediction providers, ElectricityFeeProvider.
An ElectricityFeeProvider creates forecasts for:
- electricity_fee_consumption_amt_kwh, 
- electricity_fee_consumption_percent_kwh,
- electricity_fee_feedin_amt_kwh, 
- electricity_fee_feedin_percent_kwh

The ElectricityFeeProvider forecast is updated before the ElectricityPriceProvider and the FeedInTraiffProviders are updated. Such these providers can use the newly predicted electricity fees to calculate the final electricity price and feed in tariff.

The user can select an ElectrictyFeeProvider that fits to the scheme of the system. Maybe it can be auto selected in the configuration by the geographical location of the system.

To begin with we could have the usual generic providers:
- ElectrictyFeeFixed
- ElectrictyFeeImport

and we may have a specialized electricity fee provider for Germany, starting as suggested with e.g:
- ElectricityFeeGermanyFixed

Comments welcome

NormannK:
The migration does not fully preserve existing behavior across providers. For example, `ElecPriceAkkudoktor` previously added `charges_kwh` without VAT, while the migration converts it to a fixed charge plus the default 19% VAT component. This can silently increase existing users’ configured electricity prices after upgrade.

the breaking change should be documented.

b0661 (dismissed):
Blocking this until the discussion of the best way how to integrate electricity fees into EOS has come to a conclusion. Blockage is not related to the code itself.

cornim:
Thanks @b0661, I really like this direction — a dedicated `ElectricityFeeProvider` for the more complex cases is compelling.

Before I go further: everything below is based on my currently incomplete understanding of the codebase — please correct me where I've misread how things fit together.

I'm personally a big fan of incremental changes: start with something simple that delivers value now, then extend. Concretely I'd see three steps, in order of increasing complexity:

1. **Fixed and relative charges** — absolute amounts per kWh (Netzentgelt, Konzessionsabgabe, Stromsteuer, surcharges) and percentage charges applied on a configurable basis (e.g. 19% VAT). This is what this PR already implements, as an ordered fixed/percent component list.
2. **Configured time-dependent charges** — the same components, but with per-timestamp values the user configures (e.g. HT/NT grid fees) rather than a single constant.
3. **Dynamic charges** — if/when the need arises: fees computed from a forecast (peak-load-dependent grid fees, dynamic grid tariffs), which is where a dedicated `ElectricityFeeProvider` really earns its keep.

The important constraint for me: step 1 shouldn't build anything we then have to tear down for step 2+. So I'd like to make sure the seam is shaped so 2 and 3 are additions, not rewrites.

On provider dependencies: I looked into this, and as far as I can tell there are currently no data dependencies between prediction providers — no ElecPrice/FeedInTariff/Weather/PVForecast/Load provider reads another provider's forecast at update time. The only cross-provider consumption I found is the optimization layer reading each provider's finished output from the shared container (`key_to_array` in `geneticparams.py`), and one config/logic-reuse case (Load using the separate Measurement subsystem). The comment in `prediction.py` ("providers may rely on others to be updated before") seems, in practice, to be defensive rather than backed by an actual dependency today. I'd like to preserve that independence — it keeps providers generic and avoids making `Prediction.providers` ordering load-bearing.

Given that, here's how I'd approach the steps without breaking provider independence:

- **Step 1 (this PR):** charges live in elecprice config and are applied by each price/tariff provider to its own market price. No provider reads another. The only coupling is a provider reading its own config — same as today.

- **Step 2 (time-dependent, still no cross-provider dependency):** shape the charge calculation so it operates on the market price series rather than a single scalar. Today `apply_charges` takes one scalar per timestamp; if instead it takes/returns a `pd.Series` (constant components producing today's exact results), then a future time-dependent component just carries a per-timestamp schedule and slots into the same per-component resolution loop — no signature change at the call sites, no new provider, still purely config-driven.

- **Step 3 (dynamic — introducing your `ElectricityFeeProvider`):** here the fee series would naturally come from a real provider (`ElectricityFeeProvider`), which fits the existing `DataRecord`/key machinery for free: it publishes a key (e.g. `electricity_fee_grid_percent`) just by declaring a field on its record — key registration is automatic via the Pydantic model fields, no central registry. The interesting question is where the fee gets combined with the market price, and there are two ways to do it:

  - **Design B (my preferred):** the fee provider publishes its series as an independent key, and the consumer (optimization) combines it with `elecprice_marketprice_wh` — exactly the way the optimizer already reads several providers' series side by side (`pvforecast_ac_power`, `elecprice_marketprice_wh`, `feed_in_tariff_wh`, …) in `geneticparams.py`. No provider reads another provider, so `prediction.py` ordering stays irrelevant/defensive and we keep provider independence fully intact. The tradeoff: the "market + fee" combination logic lives at the consumer, so every consumer of the electricity price has to apply it (today that's essentially the optimization parameter builder), and the static fixed/percent charges from step 1 stay in `apply_charges` while dynamic fees live in the fee-provider + consumer path. That's a slightly less tidy "one final price key," but it's the option that's consistent with how the whole codebase already works.

  - **Design A (alternative):** the `ElectricityFeeProvider` runs before elecprice, and `apply_charges` (inside the elecprice provider) reads the fee key back via `prediction.key_to_array(...)` and folds it directly into `elecprice_marketprice_wh`. The upside is a single finished consumer-price key that every consumer sees without extra logic. The tradeoff — and the reason it's not my first choice — is that this makes the elecprice provider read another provider's key during its own update. That would be the first provider→provider data dependency in the codebase and would turn `prediction.py`'s ordering into a correctness requirement rather than a defensive nicety. It's the more convenient shape but buys that convenience with exactly the coupling I'd prefer to avoid.

  Either way, step 3 is purely additive to steps 1–2: no change to the component model or the series signature from step 2, and (under Design B) no change to the other providers. So the decision can which design is preferred can be taken at a later time.

I could pull forward one thing into this PR if we think it's worth it: the scalar → series seam. Today `apply_charges` is called per-timestamp with a scalar, so a step-2 time-dependent component (which needs the timestamp index) has no natural home yet. If I reshaped `apply_charges` to take/return a `pd.Series` and apply it once per update, step 2 would later be a pure "add a component branch" touching only `elecprice.py`, not the four providers again. The counter-argument is that this enlarges this PR and the refactor is cheap and behavior-preserving to do when step 2 actually lands — so leaving the scalar API as-is now doesn't create any tear-down later. My mild preference is to leave it as-is, but I could just as easily pull it forward if you'd rather bake the seam in now.

Does the phasing make sense to you overall? And does my read that there are currently no inter-provider data dependencies match your understanding?
If so I could also create tickets for steps two and three.

cornim:
> The migration does not fully preserve existing behavior across providers. For example, `ElecPriceAkkudoktor` previously added `charges_kwh` without VAT, while the migration converts it to a fixed charge plus the default 19% VAT component. This can silently increase existing users’ configured electricity prices after upgrade.
>
> the breaking change should be documented.

@NormannK Fixed in 0661364 — rather than documenting this as a breaking change, I adapted the migration to avoid it entirely.

`_migrate_elecprice_charges` now branches on `elecprice/provider` and reproduces each provider's exact old formula instead of always emitting `fixed + 19% VAT`:

- `ElecPriceAkkudoktor` (and no provider configured, since that's the default): old code was `market + charges_kwh`, VAT never applied → migrates to a `fixed` component only.
- `ElecPriceEnergyCharts`: old code was `(market + charges_kwh) * vat_rate`, but only when `charges_kwh > 0` → migrates to `fixed` + `percent` (VAT), and only when `charges_kwh > 0`; otherwise no `charges` component is created (matching the old "price unchanged" behavior).
- `ElecPriceFixed` / `ElecPriceTibber` / `ElecPriceImport`: these never read `charges_kwh`/`vat_rate` at all in the old code → the fields are now dropped without creating any `charges` component.

This makes the migration price-neutral for existing configs. Added tests for each provider case in `tests/test_configmigrate.py` and updated the two golden migration fixtures (`eos_config_andreas_now.json`, `eos_config_minimal_now.json`) that were exercising exactly this bug (Akkudoktor and Import providers respectively) to reflect the corrected expected output. Also added a short `CHANGELOG.md` note describing the fix.

b0661:
>   The important constraint for me: step 1 shouldn't build anything we then have to tear down for step 2+. So I'd like to make sure the seam is shaped so 2 and 3 are additions, not rewrites.

I would even start more simple.  

1. Start out with the usual generic providers:

- ElectricityFeeFixed
- ElectricityFeeImport

These can be copies of the ElecPriceFixed and ElecPriceImport providers changed/ enhanced to cover the  prediction keys:

- electricity_fee_consumption_amt_kwh,
- electricity_fee_consumption_percent_kwh,
- electricity_fee_feedin_amt_kwh,
- electricity_fee_feedin_percent_kwh

No change to the current electricity and feed-in providers to start with. Should be done in a later step.

ElecPriceFixed already allows to define time windows and associated prices.  So 1. Fixed and relative charges and 2. 
Configured time-dependent charges are already covered by an enhanced copy of ElecPriceFixed. Here just the pure resulting amount should be configured (no details like Netzentgelt, Konzessionsabgabe, Stromsteuer, surcharges) to make it independent from country specifics .

2. Add a Germany specific provider

3. Change electricity providers and feed-in providers to use the new electricity fee prediction keys.

I would really make them create the full price information including fee (your Design A (alternative)) . These providers may take the fee information also from other sources and in the end the optimization would have to get an information from the provider what market price and feed-in tariff really means. This would shift responsibility to a unit (the optimization) that is not really involved in price/ feed-in tariff prediction. To have the dependency between the providers is more natural and keeps responsibility to the ones really involved.

> Does the phasing make sense to you overall? 

Shure, it will greatly ease the integration of a new concept and also the review activities,

> And does my read that there are currently no inter-provider data dependencies match your understanding?

Yes there are currently none. But the whole concept for providers was build with inter-provider dependencies in mind. There are already some comments that hint that.

>  could pull forward one thing into this PR if we think it's worth it: the scalar → series seam ... My mild preference is to leave it as-is, but I could just as easily pull it forward if you'd rather bake the seam in now.

I would leave it as it is. Adding this to Step 3 (see above) would be good, and there have to be changes to providers here anyway.

cornim:
@b0661 Given the amount of already-implemented and tested code in this PR (the ordered fixed/percent charge-component model, its wiring, and the migration logic), I'd like to propose keeping all three of your proposed steps together in this single PR rather than splitting into three separate PRs:

1. Generic `ElectricityFeeProvider`/`ElectricityFeeFixed`/`ElectricityFeeImport` producing the four fee prediction keys.
2. A Germany-specific fee provider with the structured breakdown (Netzentgelt, Konzessionsabgabe, Stromsteuer, MwSt, etc.) — reusing the underlying charge-resolution logic already built in this PR, adapted to the new provider architecture.
3. Wiring `ElecPriceAkkudoktor`/`ElecPriceEnergyCharts`/`ElecPriceFixed`/`ElecPriceTibber` (and the feed-in providers) to consume the fee provider's keys and produce the final price (Design A).

Splitting this into three separate PRs would mean discarding or significantly reworking most of what's already built here, and would also leave the first one or two PRs shipping prediction keys that nothing consumes until the last PR lands (not independently useful). It would also mean #392 stays open until that third PR lands, since the fee breakdown isn't actually reflected in any price until the wiring step is in place. Doing all three in one PR lets us land a fully working, end-to-end feature in one go, building on the logic and tests already in this PR rather than throwing them away, and close #392 as soon as it merges.

I don't want to put in this work without your go-ahead first, given the size of it — could you let me know if this approach works for you before I proceed?

b0661:
> keeping all three of your proposed steps together in this single PR rather than splitting into three separate PRs:

I ḿ fine with this.

b0661 (commented):
Now that the concept is clear the blockage can be removed.

NormannK:
@b0661 how do we proceed here? is this ready to be checked and merged?

b0661:
> @b0661 how do we proceed here? is this ready to be checked and merged?

@cornim wants to do a substantial change to this PR. Not ready to be checked and merged at the moment.

cornim:
I won't be able to work on this in the coming weeks as I have too much to do. Happy for somebody else to pick this up if they have the time. The plan should be reasonably clear I think.

b0661:
Partly superseeded by #1235. Germany specific electricity fee provider still missing.

PR #1213, opened 2026-08-01
fix: EOSdash display resampled prediction values

b0661:
Make EOSdash display resampled prediction values where resampling fits to the prediction value type.

### DIFF
--- a/src/akkudoktoreos/server/dash/prediction.py
+++ b/src/akkudoktoreos/server/dash/prediction.py
@@ -6,7 +6,7 @@ from bokeh.models import ColumnDataSource, LinearAxis, Range1d
 from bokeh.plotting import figure
 from monsterui.franken import FT, Grid, P
 
-from akkudoktoreos.core.pydantic import PydanticDateTimeDataFrame
+from akkudoktoreos.core.pydantic import PydanticDateTimeSeries
 from akkudoktoreos.server.dash.bokeh import Bokeh, bokey_apply_theme_to_plot
 from akkudoktoreos.server.dash.components import Error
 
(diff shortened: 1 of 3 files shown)


src/akkudoktoreos/server/dash/prediction.py
```diff
@@ -220,31 +222,63 @@ def Prediction(eos_host: str, eos_port: Union[str, int], data: Optional[dict] =
...
 
+    # ---------------------------------------------------------------------
+    # Merge into dataframe
+    # ---------------------------------------------------------------------
+    predictions = pd.concat(series_list, axis=1).reset_index()
+    predictions.rename(columns={"index": "date_time"}, inplace=True)
+
     # Remove time offset from UTC to get naive local time and make bokeh plot in local time
     date_time_tz = predictions["date_time"].dt.tz
     predictions["date_time"] = pd.to_datetime(predictions["date_time"]).dt.tz_localize(None)
```
NormannK, line 284: This drops the  timezone without first converting the timestamps to the configured EOS timezone.
.dt.tz_localize(None)
why do we do that?

src/akkudoktoreos/server/dash/prediction.py
NormannK, line 33: The returned data now has one point every 15 minutes, but the PV and electricity-price plots still use `BAR_WIDTH_1HOUR * 0.8` for each bar. This makes each bar 48 minutes wide, so consecutive 15-minute bars overlap substantially.

NormannK: BAR_WIDTH_15_MINUTES = 1000 * 60 * 15
width=BAR_WIDTH_15_MINUTES * 0.8

b0661: Is changed in #1214

src/akkudoktoreos/server/dash/prediction.py
b0661: We want to display the plot in the local time of the location of the system. This still does not work correctly. Especially as bokeh seems to change it's policy about time display. @cornim is working on that (see #1208).

b0661:
Superseeded by #1214

PR #1214, opened 2026-08-03
feat: add pvlib pv forecast provider

b0661:
Add a PV forecast provider that calculates the forecast using a PVLib system model
and weather forecast from the EOS weather forecast provider.

Additional module and inverter models can be easily added as the database
is build from PVLib and SAM databases and a bundled csv file.

The module model and inververt model names are provided by new endpoints
to be used in configuration.

The provider is based on the fantastic work of EMHASS. See
https://github.com/davidusb-geek/emhass/blob/master/src/emhass/forecast.py

A short description of the provider is added to the documentation.

Besides the new features there are the fixes and improvements:

* feat: improve EOSdash config page

* fix: kex_to_series for start_datetime

  Make key_to_series always start the series at start_datetime.

* fix: default provider for GENETIC and GENETIC0 optimization

  To make the default less dependent on internet servers (with API changes and
  availability issues) the default for PVForecast is set to PVForecastPVLib
  and for ElecPrice to ElecPriceFixed. The default weather provider is changed
  to OpenMeteo.

* fix: EOSdash display resampled prediction values

  Make EOSdash display resampled prediction values where  resampling fits to
  the prediction value type. Use bar width that fits to 15 minutes value samples.

* chore: add a UI hints system to EOSdash

  The UI hints system eases the definition of forms for configuration items.
  There are also forms for items in maps and lists. These forms allow to add and delete
  items to/ from  maps and lists. The forms ensure that all required fields of newly
  added items are filled.

* chore: Create an enum for valid optimization algorithms

* chore. Make config also provide the available energy management modes.

  Used for configuration hints.

* chore: Randomize default device id in configuration


src/akkudoktoreos/server/dash/uihints.py
```diff
@@ -0,0 +1,427 @@
...
+# Registry
+# ---------------------------------------------------------------------------
+
+UI_HINTS: dict[str, UiHint] = {
+    # ------------------------------------------------------------------
+    # EMS
+    # ------------------------------------------------------------------
+    "ems.mode": UiHint(
+        form="select",
+        options=["OPTIMIZATION", "PREDICTION", "None"],
```
NormannK: **Blocking:** This replaces the valid `DISABLED` enum value with `None`. `EnergyManagementMode` accepts only `DISABLED`, `PREDICTION`, and `OPTIMIZATION`, so EOSdash can submit an invalid value and no longer provides a way to disable EMS. Please derive these options from the enum or restore `DISABLED`.

src/akkudoktoreos/prediction/pvforecastpvlib.py
```diff
@@ -0,0 +1,507 @@
...
+                    f"Plane {plane_idx}: albedo set to 0.2 (was None).",
+                )
+            else:
+                df_weather["albedo"] = plane.albedo
+
+            system = PVSystem(
+                surface_tilt=plane.surface_tilt,
+                surface_azimuth=plane.surface_azimuth,
+                module_parameters=module,
+                inverter_parameters=inverter,
```
NormannK, line 368: **Blocking:** The PV system is constructed using only the selected database inverter parameters. The configured `plane.inverter_paco` limit and `plane.loss` percentage are never applied, so the resulting AC forecast can exceed the configured inverter limit and omit the configured system losses. Please apply both settings, or explicitly reject/document them as unsupported instead of silently ignoring them.

src/akkudoktoreos/server/dash/itemscard.py
```diff
@@ -0,0 +1,521 @@
...
+        wrapped = {key: wrapped}
+
+    # Outer card update state — resolved once before the inner loop
+    items_update_error = config_update_latest.get(config_name, {}).get("error")
+    items_update_value = config_update_latest.get(config_name, {}).get("value") or value
+    items_update_open = config_update_latest.get(config_name, {}).get("open") or False
+
+    # Add button: appends one new item pre-filled with model defaults
+    new_item_defaults = _item_model_defaults(item_model)
+    appended_json = json.dumps(json.dumps(items_list + [new_item_defaults]))
```
NormannK: **Blocking:** `_item_model_defaults()` omits fields without defaults, but this code immediately persists the appended object. For `HomeApplianceCommonSettings`, `consumption_wh` and `duration_h` are required, so clicking **Add item** submits an invalid configuration before the user can edit it. Please collect required values before updating, or keep the new item as an unpersisted draft until it validates.

NormannK (changes requested):
Requesting changes for three blocking functional regressions. Inline comments are anchored at the affected positions:

- `src/akkudoktoreos/server/dash/uihints.py:133` — invalid EMS mode options.
- `src/akkudoktoreos/prediction/pvforecastpvlib.py:358` — configured inverter limit and system losses are ignored.
- `src/akkudoktoreos/server/dash/itemscard.py:424` — Add item persists an invalid object for models with required fields.

The automated checks are green, but these behaviors should be corrected before merge.

src/akkudoktoreos/server/dash/uihints.py
b0661: Fixed to available modes.

src/akkudoktoreos/prediction/pvforecastpvlib.py
b0661: Fixed. Now creates warnings.

src/akkudoktoreos/server/dash/itemscard.py
b0661: Fixed: On addition of items the required fields are checked and if not provided an extra form is presented to fill it before storage.

NormannK (approved):

PR #1234, opened 2026-08-15
fix: default server settings prevent env var config

### LINKED ISSUE #1225: [ENH]: Make EOS Home Assistant add-on work also in Home Assistant Container
### Describe the issue:

EOSdash binds to 127.0.0.1 instead of 0.0.0.0, unreachable via Supervisor Ingress

After updating to version 0.3.0.dev2608080910933529 (which includes the fix from #1222), the main server crash is resolved, but the web UI is still not reachable through the Home Assistant Supervisor. EOSdash only binds to 127.0.0.1:8504 (loopback), which means it's only reachable from inside the container itself — not from the Docker network, which is how the Supervisor's ingress proxy connects to it.

Log line confirming the bind address:

`[EOSdash] __main__:run_eosdash:491 - Starting EOSdash server on 127.0.0.1:8504`

The Supervisor's ingress tries to reach the container via its Docker network IP and consistently fails:

`Ingress error: Cannot connect to host 172.30.33.4:8504 ssl:default [Connect call failed ('172.30.33.4', 8504)]`

This happens on every start, not intermittently — the port never becomes reachable from outside the container.

Expected behavior: EOSdash should bind to 0.0.0.0:8504 (or the configured host address) in the Home Assistant add-on environment so the Supervisor's ingress proxy can reach it.

Related: Follow-up to #1221 / #1222 — that fix resolved the crash, this is a separate, previously-masked issue.

```python

```

### Error message:

```shell

```

### Version information:

Akkudoktor-EOS: 0.3.0.dev2608080910933529
Environment: Home Assistant add-on; Proxmox

### LINKED ISSUE #1227: [BUG]: Default in server settings prevents configuration by environment vars
### Describe the issue:

As environment vars have lower priority than initialisation settings they never can set configuration settings with a non None default value.

```python
EOSdash IP address given by environment var does not change address used. EOSdash starts with default address localhost.
```

### Error message:

```shell

```

### Version information:

latest

b0661:
Change configuration source priorities to:

- cli
- environment vars
- dotenv settings
- config file settings
- init settings

By this the environment vars supersede any configuration var provided by the configuration file or by the initialisation with pydantic.

The test_config.py::test_computed_path was fixed to to not use the defaul env var overwrite defined by conftest.py. This seemed to indicate non working env vars, but in fact was a test setupt fault.

Besides this fix there are other fixes and changes added:

* fix: exclude computed fields when merging settings

  Pydantic may overwrite settings by values given for computed
  fields and use these values instead of re-computing the field.
  Avoid computed fields in merging settings.

* chore: improve Windows compatability of development setup

  Improve scripts to better run also on Windows. When doing path
  checks keep compatibility also to Windows pathes. A lot of
  changes to avoid the famous Windows CRLF handling and keep
  line endings to LF.

* chore: add development hint for Windows

  Windows developers should set core.autocrlf to false.

* chore: update version

Hopefullly:

- Closes: #1227
- Closes: #1225


src/akkudoktoreos/config/config.py
```diff
@@ -640,17 +644,19 @@ def lazy_file_secret_settings() -> dict:
...
         # The settings are all lazyly evaluated at instance creation time to allow for
         # runtime configuration.
         setting_sources = [
             lazy_config_cli_settings,  # Prio high
-            lazy_config_file_settings,
-            lazy_init_settings,
             lazy_env_settings,
             lazy_dotenv_settings,
+            lazy_config_file_settings,
+            lazy_init_settings,
```
NormannK, line 662: The intended priority in the PR description is `CLI > env > dotenv > config file > init`, but the actual JSON config values are loaded by `lazy_file_settings`, not `lazy_config_file_settings` (which only resolves/creates the config file and returns `{}`).

With the current ordering, `lazy_init_settings` still has higher priority than the JSON config file.
It should be.

```
setting_sources = [
    lazy_config_cli_settings,  # Prio high
    lazy_env_settings,
    lazy_dotenv_settings,
    lazy_config_file_settings,  # resolves/creates config file path
    lazy_file_settings,         # actually loads JSON config values
    lazy_init_settings,
    lazy_data_folder_path_settings,
    lazy_file_secret_settings,  # Prio low
]
```

NormannK (changes requested):

src/akkudoktoreos/config/config.py
b0661: Changed

NormannK (approved):

jogge777:
Confirmed via docker inspect on the running HA add-on container: none of the env vars EOS_SERVER__EOSDASH_HOST or EOS_SERVER__HOST are set by the add-on. Only EOS_DIR, EOS_DATA_DIR, EOS_CACHE_DIR, EOS_OUTPUT_DIR, EOS_CONFIG_DIR, and EOS_SERVER__EOSDASH_SESSKEY are present. So even with the priority fix from #1234, EOSdash still falls back to its default of 127.0.0.1, since there's no env var to take priority over the default in the first place — the add-on itself needs to set EOS_SERVER__EOSDASH_HOST=0.0.0.0 (and likely EOS_SERVER__HOST=0.0.0.0) for the Home Assistant Supervisor's ingress to be able to reach it.

jogge777:
Root cause found + workaround confirmed:
After updating to the version containing #1234, the web UI was still unreachable via the Home Assistant Supervisor's ingress. Investigated via docker exec into the running add-on container:
docker inspect <container> --format '{{range .Config.Env}}{{println .}}{{end}}' showed the add-on sets EOS_DIR, EOS_DATA_DIR, EOS_CACHE_DIR, EOS_OUTPUT_DIR, EOS_CONFIG_DIR, and EOS_SERVER__EOSDASH_SESSKEY — but no EOS_SERVER__HOST or EOS_SERVER__EOSDASH_HOST.
The persisted config file (/data/config/EOS.config.json) already had "server": {"host": "0.0.0.0", ...} — correct for the main API — but had no eosdash_host key at all, so EOSdash fell back to its Python default of 127.0.0.1.
So even with the fixed source-priority order from #1234, there's simply no env var or config value telling EOSdash to bind anywhere other than localhost — the Home Assistant add-on itself never sets this.
Workaround that fixed it for me: manually adding "eosdash_host": "0.0.0.0" next to "host": "0.0.0.0" in /data/config/EOS.config.json, then fully restarting the add-on. UI is now reachable via ingress.
Suggested proper fix: the Home Assistant add-on wrapper should set EOS_SERVER__EOSDASH_HOST=0.0.0.0 (in addition to EOS_SERVER__HOST) when starting the container, so this doesn't need to be patched manually after every add-on update.

hellebauer:
Along the lines of what jogge777 said.
I'm running Docker on Win11 (using compose).

Before the last 'fix', port 8503 would work , 8504 not at all.
Now:  x.x.x.x:8503 always gets forwarded to 127.0.0.1:8504, and nothing works.
> 
akkudoktoreos  | 2026-08-19 12:57:29.169 | INFO     | __main__:run_eos:2381 - Starting EOS server on 0.0.0.0:8503
akkudoktoreos  | 2026-08-19 12:57:29.169 | INFO     | __main__:run_eos:2383 - EOSdash will be available at 127.0.0.1:8504
> 

Temporary 'fix' [here](https://akkudoktor.net/t/fixed-eos-docker-keine-webseite-verfuegbar-weil-starting-eosdash-server-on-127-0-0-1-8504/40626/4) still works for port 8504.

PR #1235, opened 2026-08-17
feat: electricity fee provider framework and generic providers

b0661:
Add new provider class for electricity fees providers.

Add the generic providers:
- ElecFeeFixed
- ElecFeeImport

The providers provide predictions for:

- elecfee_consumption_amt_wh:
  Total fixed fee for consumed energy per Wh [amount/Wh]. This is the accumulation of all
  fixed per-Wh fees payable on "consumed energy - such as network charge, concession fee,
  and electricity charge - into a single amount.
- elecfee_consumption_percent_amt:
  Total fixed surcharge on consumed energy, given as a percentage of the monetary amount
  already charged for that energy [%]. This is the accumulation of all percentage-based
  surcharges payable on top of the consumed-energy fee - such as VAT - into a single
  percentage. This is a percentage of the fee amount, not a per-Wh rate.
- elecfee_feedin_amt_wh:
  Total fixed deduction from feed-in energy per Wh [amount/Wh]. This is the accumulation of
  all fixed per-Wh charges deducted from feed-in energy - such as metering fees or
  grid-operator handling "charges - into a single amount. Applied after the percentage-based
  deduction, i.e. it reduces the price by a flat amount per Wh rather than by a share of the
  raw price.
- elecfee_feedin_percent_amt:
  Total percentage deducted from the raw feed-in price (spot price) [%]. This is the
  accumulation of all percentage-based deductions payable on the feed-in tariff - such as a
  marketing or balancing fee retained by the aggregator - into a single percentage. It is
  applied as `raw_price * (100 - percent) / 100`, i.e. it scales down the raw price rather
  than adding a surcharge to it.

A new apply_fee() method is added to the ElecPrice and FeedInTariff base class to be used to
add the fees in a consistent way. Fees are taken from the active ElecFee provider and applied
to the raw prices given to the apply_fee() method.

The optional application of fees is added to:

- ElecPriceAkkudoktor
- ElecPriceFixed
- ElecPriceEnergyCharts
- ElecPriceSMARD
- FeedInTariffEnergyCharts
- FeedInTariffFixed
- FeedInTariffSMARD

The import providers ElecPriceImport and FeedInTariffImport do not apply fees by intentention.

The following providers currently do not handle fees defined by ElecFee:

- ElecPriceTibber
- FeedInTariffAkkudoktor
- FeedInTariffDvhubOnline
- FeedInTariffTibber

The tests for this feature are either added or existing tests are extended.

The documentation was extended for the electricity fee provider settings.

Besides this feature further improvements are added:

* feat: add SMARD quarter-hour electricty price and feed-in tariff provider

* feat: to_series method for TimeWindows and ValueTimeWindows

  Additional to to_array the time window sequence can now also produce a pandas series.
  Test have been extended to cover the series generation.

* feat: use time windows in fixed feedin tariff provider

  Feedin tariff can now be configured by time windows - not a single value.

* feat: EOSdash select for PVLib inverters and modules

  Provide PVLib inverter and module names in config selection.

* feat: EOSdash lazy select for big option sets

  Add a new form for lazy selection of big option sets. Filtering and
  generation of the option set is done server-side.

* fix: use raw data for ETS/ median prediction

  Use to raw time series data for ETS/ median prediction to avoid interference
  by e.g. dynamic grid charges.

* fix: EOSdash config drops by type only on details resolve

  Drop configuration by type and path. Prevents dropping of configuration items
  with same type and level but different path.

* fix: EOSdash configuration section closes on update

  Open section if searching or if last update touched this category — including
  updates on deeply nested sub-fields.

* chore: make elecfeefixed, elecpricefixed and feedintarifffixed warn about no windows and default to 0

  Missining configuration creates default 0 value and a warning instead of an exception.

* fix: test setup for providers

  Reset db state on each test run.

* chore: improve config option naming for elecpricefixed.

* chore: adapt elecpricefixed test to changed time_windows naming

* chore: factorized common price provider helpers to priceabc.py

  Factorized common price provider helpers to priceabc.py. Add tests for these helpers.
  Reduce/ change testing of elecpriceabc.py and feedintariffabc.py to cover
  only specifics. Rest of testing is already covered by test_priceabc.py.

* chore: update version
  
This PR is part of the #1192. It partly replaces #1212.


src/akkudoktoreos/prediction/elecfeeabc.py
```diff
@@ -0,0 +1,128 @@
...
+        json_schema_extra={"description": "List of ElecFeeDataRecord records"},
+    )
+
+    @classmethod
+    @abstractmethod
+    def provider_id(cls) -> str:
+        return "ElecFeeProvider"
+
+    def enabled(self) -> bool:
+        return self.provider_id() == self.config.elecprice.provider
```
NormannK: Shouldn't this check self.config.elecfee.provider? As written, an ElecFee* provider can only be enabled when its ID matches the electricity price provider.

src/akkudoktoreos/prediction/elecfeefixed.py
```diff
@@ -0,0 +1,191 @@
...
+        """
+        elecfee_spec: dict[str, ValueTimeWindowSequence] = {
+            "elecfee_consumption_amt_wh": self.config.elecfee.elecfeefixed.consumption_amt_kwh,
+            "elecfee_consumption_percent_amt": self.config.elecfee.elecfeefixed.consumption_percent_amt,
+            "elecfee_feedin_amt_wh": self.config.elecfee.elecfeefixed.feedin_amt_kwh,
+            "elecfee_feedin_percent_amt": self.config.elecfee.elecfeefixed.feedin_percent_amt,
+        }
+
+        for prediction_key, time_window_seq in elecfee_spec.items():
+            if time_window_seq is None or not time_window_seq.windows:
```
NormannK, line 153: Do we want to fail here? The settings document each fee component as optional, so an empty sequence should probably mean a zero fee rather than making all four components mandatory.

src/akkudoktoreos/prediction/elecfeefixed.py
```diff
@@ -0,0 +1,191 @@
...
+                align_to_interval=True,
+            )
+
+            for idx, fee in enumerate(fees):
+                if prediction_key.endswith("wh"):
+                    # Convert kWh → Wh
+                    fee = fee / 1000.0
+
+                # Store one entry per interval step.
+                current_dt = start_datetime.add(seconds=idx * interval_seconds)
```
NormannK: `to_array(..., align_to_interval=True)` may floor the start time, but these timestamps are rebuilt from the original `start_datetime`. Could this shift values when the EMS start isn't already on a 15-minute boundary? Using `to_series()` would preserve the aligned timestamps.

src/akkudoktoreos/prediction/elecfeeabc.py
b0661: Copy & paste error. Changed

src/akkudoktoreos/prediction/elecfeefixed.py
b0661: Changed to to_series.

src/akkudoktoreos/prediction/elecfeefixed.py
b0661: Empty sequences now default to 0 for elecfeefixed, elecpricefixed and feedintarifffixed. Additionally a warning is issued.

b0661:
@codex review

src/akkudoktoreos/config/configmigrate.py
```diff
@@ -62,11 +62,18 @@
...
     # - electric_vehicles
     "devices/electric_vehicles/0/initial_soc_percentage": None,
+    # elecfee
+    # =======
+    # - ElecFeeFixed
+    # - ElecFeeImport
     # elecprice
     # =========
+    "elecprice/charges_kwh": None,
+    "elecprice/vat_rate": None,
```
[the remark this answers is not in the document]
b0661: This would be usefull, but is too complex. Better the user does an informed decision on how to configure.

src/akkudoktoreos/prediction/feedintarifffixed.py
```diff
@@ -35,17 +57,59 @@ def provider_id(cls) -> str:
...
+        Args:
+            force_update: If True, forces update even if data exists.
+
+        Raises:
+            ValueError: If no time windows are configured.
+        """
+        prediction_key = "feed_in_tariff_wh"
+        time_windows_seq: ValueTimeWindowSequence = (
+            self.config.elecprice.elecpricefixed.feed_in_tariff_amt_kwh
+        )
```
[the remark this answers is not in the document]
b0661: Changed

src/akkudoktoreos/optimization/genetic/geneticparams.py
NormannK: ElecPriceFixedCommonSettings now uses elecprice_marketprice_amt_kwh, but this fallback still writes time_windows. As a result, the intended 0,288 demo price is not configured; after the retry ElecPriceFixed sees an empty sequence and defaults the electricity price to 0,0. Please change this to elecprice_marketprice_amt_kwh

src/akkudoktoreos/optimization/genetic0/genetic0params.py
NormannK: same as in geneticparams.py
ElecPriceFixed now reads elecprice_marketprice_amt_kwh, so setting time_windows here does not configure the 0,288 fallback price. The retry can therefore continue with a zero electricity price. Please update this fallback to use elecprice_marketprice_amt_kwh as well

src/akkudoktoreos/prediction/elecpriceenergycharts.py
```diff
@@ -164,17 +171,17 @@ def _parse_data(self, energy_charts_data: EnergyChartsElecPrice) -> pd.Series:
...
+            # Convert EUR/MWh to EUR/Wh
+            price_wh = price_eur_per_mwh / 1_000_000
 
             # Store in series
-            series_data.at[orig_datetime] = price_wh
+            prices_wh.at[orig_datetime] = price_wh
+
+        if self.config.elecprice.energycharts.apply_fees:
+            # Apply fees
+            prices_wh = await self.apply_fees(prices_wh)
```
NormannK: At this point fees are applied only to the prices received from Energy-Charts. Those fee-inclusive values are then used as history for ETS/median prediction, while the extrapolated future values are written without evaluating the fee schedule at their own timestamps. This gives incorrect results for time-window, weekday, or date-specific fees. Please forecast the raw market price first and apply apply_fees() to the complete API + predicted series before storing it. The same issue exists in ElecPriceAkkudoktor.

src/akkudoktoreos/prediction/elecpriceakkudoktor.py
```diff
@@ -153,24 +168,25 @@ async def _update_data(
...
-            price_wh = value.marketpriceEurocentPerKWh / (100 * 1000) + charges_wh
+            price_wh = value.marketpriceEurocentPerKWh / (100 * 1000)
 
             # Collect all values into the Pandas Series
-            series_data.at[orig_datetime] = price_wh
+            prices_wh.at[orig_datetime] = price_wh
+
+        if self.config.elecprice.akkudoktor.apply_fees:
+            # Apply fees
+            prices_wh = await self.apply_fees(prices_wh)
```
NormannK: see elecpriceenergycharts.py comment

NormannK (changes requested):

src/akkudoktoreos/prediction/elecpriceenergycharts.py
b0661: This make the whole thing complex. We now need two time series, one without fees and one with fees. Will take some time to model this.

NormannK: why do we need to store both? cant we just also assume the prices for the future?

b0661: The ETS/ median prediction has to run on raw prices to make it meaningfull. No interference by e.g. dynamic grid charge. This is a bug that is already in the current design.

NormannK: i dont fully anderstand why. but we could add this as a bug/limitation and simply continue for now.

b0661:
@codex review

src/akkudoktoreos/prediction/elecpriceenergycharts.py
```diff
@@ -112,6 +114,15 @@ def _validate_data(cls, json_str: Union[bytes, Any]) -> EnergyChartsElecPrice:
             raise ValueError(error_msg)
         return energy_charts_data
 
+    def _bidding_zone(self) -> str:
+        settings = self.config.feedintariff.energycharts
```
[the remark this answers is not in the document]
b0661: Changed

src/akkudoktoreos/prediction/elecpriceenergycharts.py
```diff
@@ -214,83 +220,149 @@ async def _update_data(
...
+
+        # Some of our data is already in the future, so we need to predict less.
+        # If we got less data we increase the prediction hours
+        covered_slots = 0
+        if self.highest_orig_datetime >= self.ems_start_datetime:
+            covered_slots = int(
+                (self.highest_orig_datetime - self.ems_start_datetime).total_seconds()
+                // resolution_seconds
+            )
+        needed_slots = self.config.prediction.hours * slots_per_hour - covered_slots
```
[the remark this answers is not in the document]
b0661: Calculation changed.

.env
```diff
@@ -11,8 +11,8 @@ DOCKER_COMPOSE_DATA_DIR=${HOME}/.local/share/net.akkudoktor.eos
 # -----------------------------------------------------------------------------
 # Image / build
 # -----------------------------------------------------------------------------
-VERSION=0.3.0.dev2608160647074704
-PYTHON_VERSION=3.13.15
```
NormannK, line 15: please lets dont downgrade again.

src/akkudoktoreos/prediction/elecpriceenergycharts.py
```diff
@@ -214,83 +220,151 @@ async def _update_data(
...
+        gross_start_datetime = self.ems_start_datetime
+
         # Determine if update is needed and how many days
         past_days = 35
+        needs_history_refresh = False
         if self.highest_orig_datetime:
-            history_series = await self.key_to_raw_series(
-                key="elecprice_marketprice_wh", start_datetime=self.ems_start_datetime
+            raw_history = await self.key_to_raw_series(
+                key="elecprice_marketprice_raw_wh", start_datetime=self.ems_start_datetime
```
NormannK: I think this is checking the wrong slice. raw_history starts at ems_start_datetime, but this block is supposed to verify that enough historical data is retained for ETS.

With a normal 24/48h forecast, the slice from ems_start_datetime onward will always be shorter than 2 * 168 * slots_per_hour, so needs_history_refresh remains true and we keep refreshing the 35-day history.

src/akkudoktoreos/optimization/genetic/geneticparams.py
```diff
@@ -311,6 +311,67 @@ async def prepare(cls) -> "Optional[GeneticOptimizationParameters]":
...
+                    interval=interval,
+                    fill_method="ffill",
+                )
+            except Exception as e:
+                logger.info(
+                    "No electricity fee data available - defaulting to demo data. Parameter preparation attempt {}: {}",
+                    attempt,
+                    e,
+                )
+                cls.config.merge_settings_from_dict(
```
NormannK: `elecfee.provider = None` is a valid fee-free configuration — `_apply_fees()` explicitly treats the absence of a fee provider as zero fees. Here, however, `prepare()` treats that same state as missing demo data and mutates the config to `ElecFeeFixed` with 0.21/kWh + 19%.

That means a user intentionally running e.g. EnergyCharts/SMARD without fees gets demo retail fees enabled just by running an optimization, changing the resulting costs/schedule.

Could we treat `elecfee.provider is None` as a valid zero-fee case here, and only install these demo defaults in an explicit demo/fallback scenario?

The same issue exists in `genetic0/genetic0params.py`.

NormannK (changes requested):

.env
b0661: Sorry, copy & paste

src/akkudoktoreos/prediction/elecpriceenergycharts.py
b0661: Good catch, don´t trust AI.

src/akkudoktoreos/optimization/genetic/geneticparams.py
b0661: I am not quite sure what is better. Explicit or implicit 0 fees. Changed it to implicit 0 fees in parameter preparation.

NormannK (approved):

PR #1237, opened 2026-08-18
fix(server): type-safe prediction import serialization

NormannK:
## What changed
- serialize `PydanticDateTimeData` and `PydanticDateTimeDataFrame` with `model_dump_json()`
- keep `json.dumps()` for plain dict payloads
- use explicit `isinstance` narrowing so mypy understands the request-body union

## Why
This follows up on #1223. The original `hasattr(..., "model_dump_json")` implementation has the right runtime intent, but it does not narrow the declared union for mypy and the long ternary also fails formatting checks.

## Validation
- compared the branch against the #1223 head and verified the functional diff is limited to the prediction import serialization block
- local checks could not be run because this environment cannot clone GitHub repositories; CI is the source of truth for pre-commit and pytest

Based on #1223.


b0661:
This is outdated for main. The changes are already applied. See https://github.com/Akkudoktor-EOS/EOS/blob/886c93c92bafe178ced39babd0a3d0ea903a525e/src/akkudoktoreos/server/eos.py#L1645.

Edit: Sorry, did not see the target branch https://github.com/Akkudoktor-EOS/EOS/tree/feat/direct-marketing-battery-grid-export

NormannK:
i cant get it to work. i might restart this in a new pr.

PR #1257, opened 2026-08-25
build(docker): harden container and shrink runtime image

r0b2g1t:
## Summary

Reworks the container build for size and security, and closes a privilege-escalation issue found while reviewing it. Image drops from **1.79 GB to 1.23 GB**.

## Security

* **Root chown followed symlinks.** `fix_data_directories_permissions()` runs as root before EOS drops privileges and walks `/data` calling `os.chown`, which follows symlinks by default. A planted symlink redirected the ownership change to its target, e.g. handing `/etc/shadow` to the `eos` user. Both calls now pass `follow_symlinks=False`.
* **Removed the baked-in `EOS_SERVER__EOSDASH_SESSKEY=s3cr3t`** from the Dockerfile and `docker-compose.yaml`. A known signing key in a public image lets anyone forge EOSdash session cookies. Unset means a random key per container.
* **Removed `seccomp:unconfined`**, which disabled Docker's default syscall filter. Added `no-new-privileges:true`, `cap_drop: ALL` and only the five caps the root phase needs (`CHOWN`, `DAC_OVERRIDE`, `FOWNER`, `SETUID`, `SETGID`).
* **Stripped setuid/setgid bits** from the base image binaries, which would otherwise let the `eos` user regain root after the drop.
* `scripts/get_version.py` is bind mounted instead of copied and `rm`'d — `RUN rm` leaves the file readable in an intermediate layer.

## Bug fix

Two `except` blocks for the base-dir `chown` interpolated `path`, the loop variable of the preceding `os.walk`. If a base dir is empty and its chown fails, `path` is unbound and the handler raises `NameError`. Now use `base_dir`.

## Image size and build

* Build toolchain (`gcc g++ gfortran`, `*-dev`) installed, used and `apt-get purge --auto-remove`'d in one layer; runtime `libopenblas0`, `liblapack3`, `libgomp1` installed manual so they survive.
* `python3-venv` installed only when the base lacks `venv` — saves a redundant Python on `python-slim`, keeps the HA `base-debian:trixie` build working.
* Eight `RUN` layers collapsed to two; pip cache is a BuildKit cache mount.
* Removed `RUN echo "Building ... $PYTHON_VERSION"` — `ARG` is out of scope after `FROM`, so it printed empty and cost a layer.
* Dropped `wheel`; deduplicated `mkdir`/`chown`; added `tests/`, `docs/` and tool caches to `.dockerignore`.
* Added a `HEALTHCHECK` on `/v1/health` using stdlib `urllib` (no `curl` needed).

## Deliberately unchanged

**EOS still starts as root and drops via `--run_as_user eos`.** A static `USER eos` is the stronger default but breaks the documented bind-mount deployment: a container starting as uid 999 cannot chown a host-owned `/data`. Moving to a named volume would remove the root phase entirely, but needs a data migration and its own PR. This PR shrinks the root window instead.

The editable install is kept: the version is hashed from the live source tree at import, so `src/` must exist at runtime.

## Known follow-up

`drop_root_privileges()` calls `pwd.getpwuid(os.geteuid()).pw_name` unguarded, so `docker run --user 1000:1000` crashes with `KeyError` on a uid absent from `/etc/passwd`. Left out to keep this PR scoped; worth its own issue.

## Not done

No multi-stage build — the dual base (`BUILD_FROM` HA Debian vs `python-slim`) makes venv interpreter provenance differ between stages. No separate dependency layer — `version.txt` needs `src/`, and `pyproject.toml` reads the version from `version.txt`.

## Testing

New test `test_chown_never_follows_symlinks` asserts every `os.chown` call passes `follow_symlinks=False`; confirmed it fails when the fix is reverted.

Verified locally on the `python-slim` variant:

| Check | Result |
|---|---|
| Image size | 1.79 GB → 1.23 GB |
| setuid/setgid binaries | 10 → 0 |
| PID 1 after drop | uid 999, `CapEff: 0000000000000000` |
| Default seccomp profile | healthy — `unconfined` was unnecessary |
| `/v1/health`, `HEALTHCHECK` | `alive` / `healthy` |
| `/data/config` writable by `eos` | yes |
| Compilers, `scripts/`, SESSKEY in image | absent |
| `docker compose config` | valid |

Only the new test ran locally; the full suite starts real servers and should run in CI. The Home Assistant add-on build path was not built locally and needs a CI build, since the setuid strip now touches that base too.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -13,3 +13,8 @@ Makefile
 NOTICE
 README.md
 .venv/
+tests/
+docs/
+.pytest_cache/
+.mypy_cache/
+.ruff_cache/
--- a/Dockerfile
+++ b/Dockerfile
@@ -90,7 +90,10 @@ RUN --mount=type=bind,source=scripts/get_version.py,target=${EOS_DIR}/scripts/ge
     && apt-get purge -y --auto-remove \
         gcc g++ gfortran \
         libopenblas-dev liblapack-dev \
-    && rm -rf /var/lib/apt/lists/*
+    && rm -rf /var/lib/apt/lists/* \
+    # EOS never needs setuid/setgid binaries. Stripping them keeps the unprivileged
+    # eos user from regaining root after the server drops privileges.
+    && find / -xdev \( -perm -4000 -o -perm -2000 \) -type f -exec chmod a-s {} +
 
 ENTRYPOINT []
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,71 +25,84 @@ ENV EOS_OUTPUT_DIR="${EOS_DATA_DIR}/output"
 ENV EOS_CONFIG_DIR="${EOS_DATA_DIR}/config"
 ENV MPLCONFIGDIR="${EOS_DATA_DIR}/mplconfigdir"
 
-# Overwrite when starting the container in a production environment
-ENV EOS_SERVER__EOSDASH_SESSKEY=s3cr3t
+# EOS_SERVER__EOSDASH_SESSKEY is deliberately NOT set here.
+# Baking a session signing key into a public image lets anyone forge EOSdash
+# session cookies. Unset means EOSdash generates a random key per container.
+# Set it via the environment only if sessions must survive a restart.
 
 # Set environment variables to reduce threading needs
 ENV OPENBLAS_NUM_THREADS=1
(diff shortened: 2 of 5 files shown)


b0661:
Any reason fo rclosing this? Looked good. Just unsure If there are any implications when running as a Home Assistent app.

PR #1261, opened 2026-08-29
build: multi-stage Docker image with uv and BuildKit caching

r0b2g1t:
The container image was a single stage that kept the whole build toolchain (`gcc`, `g++`, `gfortran`, the `-dev` libraries) in the final image, installed dependencies with `pip` while the rest of the project has moved to `uv`, and declared a BuildKit syntax version without using any of its caching features.

This splits the Dockerfile into a `builder` stage and a `runtime` stage that share the same base image. The builder compiles everything and installs dependencies with `uv` from `uv.lock` (`--frozen`, so builds are reproducible); the runtime stage only gets the `virtualenv` plus the shared libraries it needs at run time. Dependency installation sits in its own layer keyed on `pyproject.toml/uv.lock`, so editing source no longer triggers a full reinstall, and a cache mount keeps the `uv` download cache between builds. The standalone image drops from roughly 1.8 GB to 1.2 GB.

The project install stays editable on purpose: `akkudoktoreos.core.version` computes the version at import time and requires the `src/akkudoktoreos` layout to be present, so the runtime stage copies `src/` back in. `python3` is installed explicitly in both stages because the Home Assistant base image does not ship it.

Also fixes the `io.hass.version` label, which was the literal string `VERSION`, adds the standard OCI image labels, and adds a `HEALTHCHECK` against `/v1/health`.

**Testing**:

Built and ran both paths locally — standalone (`python:slim`) and the Home Assistant add-on base (`amd64-base-debian:trixie`). Both containers report healthy, EOSdash starts, the privilege drop to the `eos` user works, and the image runs with a read-only root filesystem.


Dockerfile
```diff
@@ -43,53 +89,29 @@ ENV PATH="$VENV_PATH/bin:$PATH"
...
-RUN pip install --no-cache-dir -e .
-
 ENTRYPOINT []
 
-EXPOSE 8504
 EXPOSE 8503
+EXPOSE 8504
+
+HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
+    CMD python -c "import urllib.request; urllib.request.urlopen('http://127.0.0.1:8503/v1/health')" || exit 1
```
NormannK: The server port is configurable through `EOS_SERVER__PORT` (the configuration tests exercise values other than 8503), but this healthcheck always probes port 8503. A standalone container configured for another internal port will therefore be marked unhealthy even though EOS is running. Could this build the URL from `EOS_SERVER__PORT`, with 8503 as the fallback, or otherwise make the container port fixed?

Dockerfile
```diff
@@ -1,19 +1,65 @@
...
+# Install the project itself. Editable, because akkudoktoreos.core.version
+# requires the src/akkudoktoreos layout at runtime; the runtime stage copies
+# src/ back in.
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv sync --frozen --no-dev
+
+
+FROM ${BUILD_FROM:-python:${PYTHON_VERSION}-slim} AS runtime
+
+ARG BUILD_VERSION=VERSION
```
NormannK: This default still leaves the version labels as the literal `VERSION` whenever the caller does not supply `BUILD_VERSION`. Neither `docker-compose.yaml` nor `.github/workflows/docker-build.yml` currently passes it, so the claimed fix does not cover those build paths (`org.opencontainers.image.version` may be overridden by metadata-action, but `io.hass.version` is not). Please pass `BUILD_VERSION` from those callers or provide a real fallback.

NormannK (changes requested):
The multi-stage/uv change looks sound overall, but these two configuration paths need to be corrected before merge.

Dockerfile
r0b2g1t: Good catch. `HEALTHCHECK` runs through `/bin/sh` at container start, so the value is available at runtime — I've changed the check to read `EOS_SERVER__PORT` with `8503` as the fallback and added an explicit `timeout=3` so the probe can't outlive `--timeout=5s`.

I deliberately did not add `ENV EOS_SERVER__PORT=8503` to the image: env vars outrank `EOS.config.json` in EOS's config precedence, so a baked-in default would silently override anyone who configures the port through the config file. The remaining gap is exactly that case — a port set only in `EOS.config.json` won't be picked up by the healthcheck. If you'd prefer full coverage there, the alternative is to have the check load the EOS config instead of the env var; happy to do that, it's just a heavier probe. For the compose path the container-internal port is fixed at `8503` anyway ("`${EOS_SERVER__PORT}:8503`" only maps the host side), so this only affects hand-rolled docker run invocations.

Dockerfile
r0b2g1t: Good catch — the indirection didn't actually fix anything for standalone builds.

Changed the default from the literal `VERSION` to dev, so an unwired local build gets an honest marker instead of a bogus version string. docker-compose.yaml now passes `BUILD_VERSION: "${VERSION}"` (same `.env` var already used for the image tag), so compose builds get the real version. The Home Assistant builder passes `BUILD_VERSION` on its own, so add-on builds were always fine.

I also dropped `org.opencontainers.image.version` from the `Dockerfile` — the CI workflow already sets it through docker/metadata-action, so baking a default there was both redundant and the misleading part. `io.hass.version` only matters for the HA add-on build anyway, which supplies the arg. That leaves the CI workflow untouched.

NormannK:
Both original fixes look good. One remaining edge case is a port configured only through `/data/config/EOS.config.json`: the server will use that port, but the current healthcheck falls back to `8503` because it only checks `EOS_SERVER__PORT`.

A small stdlib-only probe module would preserve the existing configuration behavior without importing the full application on every healthcheck:

```python
# src/akkudoktoreos/server/container_healthcheck.py
import json
import os
from pathlib import Path
from urllib.request import urlopen

DEFAULT_PORT = 8503


def effective_port() -> int:
    if value := os.getenv("EOS_SERVER__PORT"):
        return int(value)

    config_file = Path(os.getenv("EOS_CONFIG_DIR", "/data/config")) / "EOS.config.json"
    try:
        config = json.loads(config_file.read_text(encoding="utf-8"))
    except (FileNotFoundError, OSError, json.JSONDecodeError):
        return DEFAULT_PORT

    server = config.get("server") or {}
    value = server.get("port", DEFAULT_PORT)
    port = DEFAULT_PORT if value is None else int(value)
    if not 1 <= port <= 65535:
        raise ValueError(f"Invalid server port: {port}")
    return port


def main() -> int:
    try:
        port = effective_port()
        with urlopen(f"http://127.0.0.1:{port}/v1/health", timeout=3) as response:
            return 0 if response.status == 200 else 1
    except (OSError, ValueError):
        return 1


if __name__ == "__main__":
    raise SystemExit(main())
```

The Dockerfile can then use exec form:

```dockerfile
HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
    CMD ["python", "-m", "akkudoktoreos.server.container_healthcheck"]
```

Separately, if the published CI image should also carry an accurate `io.hass.version`, the build step can pass the metadata-generated version:

```yaml
build-args: |
  BUILD_VERSION=${{ steps.meta.outputs.version }}
```

That leaves `dev` as the honest fallback for direct local builds while giving Home Assistant, Compose, and CI explicit versions.

NormannK (approved):

NormannK:
@b0661 could you please check? i think i solved the problems.

Issue #1274, opened 2026-09-04
[BUG]: Config path test fails on macOS due to /var alias


dr-dimitri:
### Describe the bug

`tests/test_config.py::test_get_config_file_path` fails on macOS because the production code resolves configuration directories while the test compares the returned path with an unresolved `TemporaryDirectory` path.

macOS exposes temporary directories through `/var/folders/...`, while `Path.resolve()` canonicalizes the same location to `/private/var/folders/...`. The paths identify the same file, but direct `Path` equality fails.

### Steps to reproduce

1. Check out the current `main` branch on macOS.
2. Install the development environment.
3. Run:

   ```shell
   pytest -q tests/test_config.py::test_get_config_file_path
   ```

### Actual behavior

The assertion for `EOS_DIR` fails with paths equivalent to:

```text
actual:   /private/var/folders/.../EOS.config.json
expected: /var/folders/.../EOS.config.json
```

`ConfigEOS._get_config_file_path()` calls `Path.resolve()` for directories read from `EOS_DIR` and `EOS_CONFIG_DIR`. The test helper builds expected paths directly from `Path(temp_dir)` without applying the same normalization.

### Expected behavior

The test should pass on macOS and Linux while still verifying that `ConfigEOS._get_config_file_path()` selects the correct directory and filename.

### Suggested implementation requirements

- Keep the production path-normalization behavior unchanged.
- Normalize the expected temporary paths before comparing them, or compare paths using another platform-safe canonical representation.
- Cover every assertion in `test_get_config_file_path` that derives an expected path from the temporary directory.
- Verify the focused test on macOS and ensure the existing configuration tests remain valid on Linux.

### Context

This was found while running the full test suite for #1188 and is intentionally not fixed as part of that unrelated feed-in tariff change.

Observed on current `main` at commit `d5df763c6c7febc839325f1e9c2b3b3a588134cd` with Python 3.13 on macOS.

Issue #1273, opened 2026-09-04
[BUG]: ElecPriceEnergyCharts skips refresh when final price intervals are missing


dr-dimitri:
### Describe the issue

`ElecPriceEnergyCharts._update_data()` can treat an incomplete final day of quarter-hourly
electricity prices as complete and skip the required API refresh.

The update decision currently calculates an expected last timestamp of 23:00 for the current day
before 14:00, or 23:00 for the following day from 14:00 onward, and compares that value directly to
`highest_orig_datetime`:

```python
hours_ahead = 23 if now.time() < pd.Timestamp("14:00").time() else 47
end = midnight + pd.Timedelta(hours=hours_ahead)

# ...
needs_update = end > self.highest_orig_datetime
```

Energy-Charts timestamps identify the **start** of a price interval. With 15-minute prices, a last
timestamp of 23:00 only covers 23:00–23:15. The 23:15, 23:30, and 23:45 intervals are still missing,
but `end > highest_orig_datetime` evaluates to `False` and no refresh is requested.

This is the electricity-price equivalent of the feed-in tariff bug reported in #1188. It was found
while reviewing the implementation for that issue and was deliberately not included in that fix to
keep its scope focused.

### Reproduction

1. Configure `ElecPriceEnergyCharts` and retain enough historical raw price data to avoid the
   history-repair branch.
2. Before 14:00 local time, set the latest original quarter-hourly price timestamp to 23:00 of the
   current day. Alternatively, after 14:00, use 23:00 of the following day.
3. Run `_update_data(force_update=False)`.
4. Observe that `_request_forecast()` is not called even though three expected intervals are
   missing.

### Expected behavior

The provider should compare the exclusive end of actual coverage with the expected publication
boundary:

```python
coverage_end = highest_orig_datetime + source_interval
needs_update = coverage_end < required_midnight
```

Before 14:00, `required_midnight` is the end of the current local calendar day. From 14:00 onward,
it is the end of the following local calendar day. Calendar-day calculation should remain correct
across daylight-saving transitions.

### Suggested implementation and tests

- Determine the source resolution from the retained raw history, as the provider already does.
- Treat the newest timestamp as an interval start and add one source interval before comparing
  coverage.
- Preserve the existing forced-update, history-repair, raw-price, fee-application, and forecasting
  behavior.
- Add regression cases for:
  - 15-minute data ending at 23:00: incomplete;
  - 15-minute data ending at 23:45: complete;
[...]

PR #1275, opened 2026-09-04
fix: honor Energy-Charts feed-in interval coverage

### LINKED ISSUE #1188: [BUG]: FeedInTariffAkkudoktor does not fill all 15 minutes slots
### Describe the issue:

From @NormannK by pm:

In src/akkudoktoreos/prediction/feedintariffenergycharts.py, the script checks whether new data needs to be fetched:

midnight = now.normalize()
hours_ahead = 23 if now.time() < pd.Timestamp(“14:00”).time() else 47
end = midnight + pd.Timedelta(hours=hours_ahead)

# ...

end > self.highest_orig_datetime

With 15-minute intervals, an error occurs if the stored data ends at 11:00 PM.

The timestamp 11:00 PM represents only the interval from 11:00 PM to 11:15 PM. The intervals at 11:15 PM, 11:30 PM, and 11:45 PM are still missing.

Since `end` is also set to 11:00 PM, the comparison:

end > self.highest_orig_datetime

returns `False`. As a result, no update is performed, even though the data for the day is not yet complete.

The end time should be checked against the end of the last available interval. For 15-minute data, the day is not complete until the last timestamp is 11:45 p.m., thus covering data up to midnight.
For example:
coverage_end = self.highest_orig_datetime + pd.Timedelta( seconds=resolution_seconds )
needs_update = coverage_end < next_midnight

```python

```

### Error message:

```shell

```

### Version information:

#1187

dr-dimitri:
## Summary

- Treat Energy-Charts timestamps as interval starts when deciding whether the published feed-in tariff horizon is complete.
- Compare the end of the final source interval with local midnight instead of comparing the final interval start.
- Preserve hourly behavior while correctly refreshing incomplete 15-minute data before and after the 14:00 publication cutoff.
- Add regression coverage for quarter-hourly and hourly data, both publication horizons, and a Europe/Berlin DST transition.

Closes #1188.

## Focused review

The post-implementation review checked interval-boundary semantics, compatibility with hourly source data, the exact 14:00 cutoff, calendar-day handling across DST, and that the change remains limited to the Energy-Charts feed-in tariff provider.

Two independent pre-existing issues discovered during review and full-suite verification were reported separately and were not fixed in this pull request:

- #1273: the analogous interval-coverage problem in `ElecPriceEnergyCharts`
- #1274: a macOS `/var` versus `/private/var` path-alias failure in a configuration test

## Testing

- `pytest -q tests/test_feedintariffenergycharts.py` — 17 passed
- Targeted `pre-commit run --files ...` — all applicable checks passed, including isort, Ruff lint/format, and mypy
- Full local suite — 1511 passed, 15 skipped, 5 failed, and 19 errors; the remaining results are unrelated to this change and stem from the macOS path alias reported in #1274, generated-document/version differences, and sandbox-denied process inspection in server/system tests

This pull request was implemented and reviewed by OpenAI Codex, a large language model (LLM).

### DIFF
--- a/src/akkudoktoreos/prediction/feedintariffenergycharts.py
+++ b/src/akkudoktoreos/prediction/feedintariffenergycharts.py
@@ -55,6 +55,25 @@ class FeedInTariffEnergyCharts(FeedInTariffProvider):
         """Return the unique identifier for the Energy-Charts feed-in tariff provider."""
         return "FeedInTariffEnergyCharts"
 
+    def _has_complete_published_horizon(
+        self, *, now: pd.Timestamp, resolution_seconds: int
+    ) -> bool:
+        """Return whether stored source data covers all currently published intervals.
+
+        Energy-Charts timestamps identify interval starts. The actual coverage therefore ends one
+        source interval after ``highest_orig_datetime``. Before 14:00, prices through the end of
+        the current day are expected; from 14:00 onward, the following day is expected as well.
+        """
+        if self.highest_orig_datetime is None:
+            return False
+
+        published_days = 1 if now.hour < 14 else 2
+        required_coverage_end = now.normalize() + pd.DateOffset(days=published_days)
+        coverage_end = pd.Timestamp(self.highest_orig_datetime) + pd.Timedelta(
+            seconds=resolution_seconds
+        )
+        return coverage_end >= required_coverage_end
+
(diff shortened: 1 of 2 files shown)


NormannK:
very good thanks!

Issue #1277, opened 2026-09-05
[BUG]: to_datetime shifts naive strings with omitted or fractional seconds


dr-dimitri:
## Problem

`to_datetime()` interprets equivalent timezone-naive strings differently depending on their precision. Inputs with `HH:mm:ss` retain their wall-clock time in the requested/default timezone, but inputs with `HH:mm` or fractional seconds are interpreted as UTC and then shifted. Near midnight, this silently changes the calendar date.

## Reproduction

On `main` at `940aa1021fdcd10f9fc3a36295a08ca8b21c4155`:

```python
from akkudoktoreos.utils.datetimeutil import to_datetime

for value in (
    "2026-01-15 23:45",
    "2026-01-15 23:45:00",
    "2026-01-15T23:45:00.000",
):
    print(to_datetime(value, in_timezone="Europe/Berlin"))
```

Actual:

```text
2026-01-16 00:45:00+01:00
2026-01-15 23:45:00+01:00
2026-01-16 00:45:00+01:00
```

Expected: all three represent `2026-01-15 23:45:00+01:00`.

The same issue occurs with the `T` separator and with Pendulum's default local timezone set to Europe/Berlin. In summer, the incorrect shift is two hours.

## Affected area and cause

`src/akkudoktoreos/utils/datetimeutil.py`, `to_datetime()`, around lines 924-952.

The explicit format list handles seconds but not omitted/fractional seconds. Remaining strings reach `pendulum.parse(date_input)` without a timezone argument; that parser defaults timezone-naive inputs to UTC. The later timezone conversion shifts their wall-clock time.

The docstring promises local-time interpretation for strings without timezone information. Existing tests establish local interpretation for the supported seconds format. The timezone-conversion example in the docstring should be reconciled with that behavior during this fix.

## Acceptance criteria

- Equivalent naive strings with minutes, seconds, and fractional seconds produce equivalent instants in the requested/default timezone.
- Support both space and `T` date/time separators.
- Add deterministic tests for explicit Europe/Berlin, default local Europe/Berlin, winter and summer dates, and a time near midnight.
- Preserve fractional-second precision.
- Explicit `Z` or numeric UTC offsets retain their instant and continue converting correctly to the target timezone.
- Preserve Unix timestamp behavior and existing supported date formats.
- Keep the implementation focused on this parsing inconsistency and update directly affected documentation examples.
- Run the datetime utility regression tests.

[...]

Issue #1276, opened 2026-09-05
[BUG]: make mypy and the pre-commit hook use inconsistent typing environments


dr-dimitri:
## Problem

The documented `make mypy` command and the pre-commit mypy hook produce materially different results on the same checkout. The hook succeeds, while the development-environment command reports 961 errors in 125 files. A passing hook therefore does not establish that the documented typing command passes.

Observed on upstream `main` at `940aa1021fdcd10f9fc3a36295a08ca8b21c4155`, Python 3.13.15 and mypy 2.3.1 on macOS. The same 961 error messages were confirmed with the #1273 changes and with the unchanged upstream source/test files supplied through mypy's `--shadow-file` option.

## Reproduction

```shell
uv sync --frozen --extra dev
make mypy
uv run pre-commit run mypy --all-files
```

Actual: `make mypy` fails; the pre-commit hook passes.

Expected: supported local and CI typing entry points use a consistent, reproducible environment and policy, and report the same diagnostics for the same source revision.

## Affected configuration and cause

- `Makefile` runs `uv run mypy` in the project's fully installed development environment.
- `.pre-commit-config.yaml` uses an isolated `pre-commit/mirrors-mypy` environment with only the listed stub/tool dependencies.
- The mirror's own hook definition adds `--ignore-missing-imports --scripts-are-modules`, which the Makefile command does not use.
- The isolated environment lacks typed runtime dependencies such as Pydantic and Pendulum. Ignored missing imports become `Any`, hiding errors visible in the development environment. Transitive dependencies can also differ from `uv.lock` (observed NumPy 2.5.2 in the hook versus 2.4.6 in the development environment).
- Both entry points select `src` and `tests` through `pyproject.toml`; file selection does not explain the discrepancy.

## Acceptance criteria

- Establish a single documented typing policy, using consistent arguments and reproducible dependency versions for Makefile, pre-commit, and CI.
- Ensure the typing environment includes the runtime type information the selected policy is intended to check.
- Verify the entry points report equivalent diagnostics on the same checkout, including from a fresh environment/cache.
- Demonstrate that a deliberate type error involving a typed project dependency is caught by all entry points, then remove that probe.
- Document any existing typing debt explicitly; do not silently obtain parity by disabling additional checks or ignoring all imports.
- Keep any broad existing typing cleanup separate and coordinated with the configuration fix.

Found during validation for #1273 and intentionally left out of that provider fix.

PR #1278, opened 2026-09-05
fix: honor Energy-Charts electricity price interval coverage

### LINKED ISSUE #1273: [BUG]: ElecPriceEnergyCharts skips refresh when final price intervals are missing

dr-dimitri:
Quarter-hourly Energy-Charts prices ending at 23:00 previously suppressed the next refresh even though the final three intervals were missing. `ElecPriceEnergyCharts` now adds the source interval to the newest timestamp and compares that coverage end with local midnight: the end of today before 14:00, or the end of tomorrow from 14:00 onward. Calendar-day boundaries also handle both DST transitions correctly.

Closes #1273.

## Validation

- Added 11 deterministic regression scenarios, each run under UTC and Europe/Berlin host timezones (22 cases), exercising the actual update/fetch path, stored values, hourly compatibility, the exact 14:00 cutoff, and spring/fall DST transitions.
- Request-date assertions account for the existing host-timezone formatting, while publication boundaries stay in the configured provider timezone. Returned timestamps are compared as pandas timestamps across timezones.
- Related provider suites: 72 passed, 1 skipped (`tests/test_elecpriceenergycharts.py`, `tests/test_feedintariffenergycharts.py`, `tests/test_priceabc.py`, and `tests/test_elecpriceabc.py`, with `--check-config-side-effect`). The existing resampling test needs access to the real Energy-Charts API.
- All applicable pre-commit checks passed.
- `make mypy` still fails with 961 existing errors in 125 files. Comparing normalized diagnostics with unchanged upstream source/test files using `--shadow-file` showed no added or removed errors. The isolated pre-commit mypy hook passes; this environment discrepancy is reported in #1276.
- Focused post-implementation review found no regressions or scope issues. Forced refresh, history repair, raw prices, fees, and forecasting retain their existing behavior.

The independent datetime parsing bug discovered during testing is reported in #1277 and is not part of this change.

This pull request was created, implemented, and reviewed by OpenAI Codex, a large language model (LLM).

### DIFF
--- a/src/akkudoktoreos/prediction/elecpriceenergycharts.py
+++ b/src/akkudoktoreos/prediction/elecpriceenergycharts.py
@@ -98,6 +98,25 @@ class ElecPriceEnergyCharts(ElecPriceProvider):
         """Return the unique identifier for the Energy-Charts provider."""
         return "ElecPriceEnergyCharts"
 
+    def _has_complete_published_horizon(
+        self, *, now: pd.Timestamp, resolution_seconds: int
+    ) -> bool:
+        """Return whether stored source data covers all currently published intervals.
+
+        Energy-Charts timestamps identify interval starts. The actual coverage therefore ends one
+        source interval after ``highest_orig_datetime``. Before 14:00, prices through the end of
+        the current day are expected; from 14:00 onward, the following day is expected as well.
+        """
+        if self.highest_orig_datetime is None:
+            return False
+
+        published_days = 1 if now.hour < 14 else 2
+        required_coverage_end = now.normalize() + pd.DateOffset(days=published_days)
+        coverage_end = pd.Timestamp(self.highest_orig_datetime) + pd.Timedelta(
+            seconds=resolution_seconds
+        )
+        return coverage_end >= required_coverage_end
+
     @classmethod
     def _validate_data(cls, json_str: Union[bytes, Any]) -> EnergyChartsElecPrice:
         """Validate Energy-Charts Electricity Price forecast data."""
@@ -211,11 +230,8 @@ class ElecPriceEnergyCharts(ElecPriceProvider):
 
         The final mapped and processed data is inserted into the sequence as `ElecPriceDataRecord`.
         """
-        # New prices are available every day at 14:00
+        # Tomorrow's prices are available every day at 14:00.
         now = pd.Timestamp.now(tz=self.config.general.timezone)
-        midnight = now.normalize()
-        hours_ahead = 23 if now.time() < pd.Timestamp("14:00").time() else 47
-        end = midnight + pd.Timedelta(hours=hours_ahead)
 
         if not self.ems_start_datetime:
             raise ValueError(f"Start DateTime not set: {self.ems_start_datetime}")
@@ -254,8 +270,10 @@ class ElecPriceEnergyCharts(ElecPriceProvider):
                 elif force_update:
                     # Use default start date in case of forced update
                     needs_update = True
-                elif end > self.highest_orig_datetime:
-                    # We got enough history, but still not enough data to prediction end
+                elif not self._has_complete_published_horizon(
+                    now=now, resolution_seconds=resolution_seconds
+                ):
+                    # We have enough history, but not every expected source interval.
                     start_datetime = gross_start_datetime
                     needs_update = True
         else:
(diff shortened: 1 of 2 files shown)


src/akkudoktoreos/prediction/elecpriceenergycharts.py
```diff
@@ -254,8 +270,10 @@ async def _update_data(self, force_update: Optional[bool] = False) -> None:
                 elif force_update:
                     # Use default start date in case of forced update
                     needs_update = True
-                elif end > self.highest_orig_datetime:
-                    # We got enough history, but still not enough data to prediction end
+                elif not self._has_complete_published_horizon(
+                    now=now, resolution_seconds=resolution_seconds
```
NormannK, line 274: Could we infer the interval used for the coverage check from source data ending at `highest_orig_datetime`?

Currently, `resolution_seconds` comes from history before `ems_start_datetime`. If that history is hourly but the latest published day uses 15-minute intervals, a last timestamp of 23:00 would be treated as covering through midnight, skipping the missing final intervals.

Keep the existing resolution for the history-length check, then infer the coverage resolution separately from raw data through `highest_orig_datetime`, excluding the predicted tail. For example, replace this `elif` branch (including its body) with:

```python
                else:
                    # Infer coverage resolution from the latest source data.
                    # Exclude extrapolated values after highest_orig_datetime.
                    source_series = await self.key_to_raw_series(
                        key="elecprice_marketprice_raw_wh",
                        end_datetime=to_datetime(self.highest_orig_datetime).add(seconds=1),
                    )
                    source_resolution_seconds = self._resolution_seconds(source_series)

                    if not self._has_complete_published_horizon(
                        now=now,
                        resolution_seconds=source_resolution_seconds,
                    ):
                        start_datetime = gross_start_datetime
                        needs_update = True
```

NormannK (commented):
The exclusive-end coverage calculation addresses the reported uniform-resolution bug, and the calendar-day boundary calculation handles DST correctly. One additional robustness improvement is suggested inline for mixed hourly/quarter-hourly source data. The suggested change has not been applied or tested against the full provider suite.

Issue #1279, opened 2026-09-05
[BUG]: Energy-Charts coverage uses stale resolution after source interval changes


dr-dimitri:
## Problem

`ElecPriceEnergyCharts` can still skip a required refresh when retained history and the latest published prices use different source resolutions. This is a follow-up to #1273 and the review of merged PR #1278:

https://github.com/Akkudoktor-EOS/EOS/pull/1278#discussion_r3940805727

The coverage check correctly adds one interval to `highest_orig_datetime`, but its interval comes from history ending before `ems_start_datetime`. If that history is hourly and the latest prices are quarter-hourly, 23:00 is incorrectly treated as covering through midnight.

## Reproduction

1. Retain more than two weeks of hourly raw electricity prices before the EMS start.
2. Store quarter-hourly source prices for the current day through 23:00 and set `highest_orig_datetime` accordingly.
3. Before 14:00 in the provider timezone, call `_update_data(force_update=False)`.
4. The refresh is skipped even though 23:15, 23:30, and 23:45 are missing. After 14:00, the equivalent problem affects the following published day.

Reading source data through `highest_orig_datetime` fixes a full-day transition. However, the shared `_resolution_seconds()` computes a median over up to 96 time differences. A short trailing run, such as five new quarter-hourly points after hourly history, still yields an hourly median and can trigger the same incorrect coverage decision.

## Expected behavior and acceptance criteria

- Infer the coverage interval from the most recent original source data, bounded at `highest_orig_datetime` and excluding the extrapolated tail.
- Recognize a short consistent new cadence, including five consecutive quarter-hourly points after hourly history. Document how ambiguous or insufficient evidence falls back; a single exceptional gap should not redefine an otherwise stable cadence.
- Preserve the separate resolution and threshold used for the history-length check. Do not change the shared forecasting, resampling, fee calculation, or source-resolution policy used by other providers.
- Add deterministic update-path regressions for hourly-to-quarter-hourly transitions: 23:00 requires a fetch, while 23:45 is complete. Cover both publication horizons, short trailing runs, and quarter-hourly-to-hourly compatibility.
- Verify that extrapolated timestamps cannot influence coverage resolution, including when their spacing differs from the latest original prices.
- Preserve forced updates, empty/insufficient-history repair, and fallback after fetch failures.
- Keep the existing UTC/Berlin host-timezone, exact 14:00 cutoff, and DST tests passing.
[...]

PR #1280, opened 2026-09-05
fix: infer Energy-Charts coverage from recent source intervals

NormannK:
## Problem

After #1278, the coverage check still takes its interval size from history before `ems_start_datetime`. When hourly history is followed by quarter-hourly published prices, a final timestamp of 23:00 can incorrectly count as covering through midnight. The reverse transition can cause unnecessary refreshes.

## Changes

- Keep the existing resolution and length check for forecasting history.
- Infer coverage resolution separately from the final 24 hours ending at `highest_orig_datetime`, excluding the predicted tail stored under the same raw key.
- Bound the source sample to 24 hours so older, finer intervals do not dominate the existing median estimator when the latest published day is hourly.
- Expand the existing coverage regression matrix to 44 cases with independently varied history resolution and an opposing-resolution predicted tail. This retains the publication cutoff, UTC/Berlin host timezone, stored-value, and DST assertions.

Follow-up to the review in #1278: https://github.com/Akkudoktor-EOS/EOS/pull/1278#discussion_r3940805727

## Validation

- Expanded coverage tests against unchanged merged #1278 code: 12 failed, 32 passed, reproducing missed and unnecessary refreshes.
- Final coverage tests: **44 passed**, with `--check-config-side-effect`.
- Related provider suites during implementation: 89 passed, 1 skipped, 1 deselected; the 4 remaining failures identified the reverse-transition median issue and all pass in the final 44-case run. The live-API resampling test was deselected. The full related suites were not rerun after the final 24-hour bound adjustment.
- All applicable pre-commit checks passed, including isort, Ruff, formatting, and mypy.

The existing median estimator remains heuristic if resolution changes within the final source-day sample; this change targets differing historical and published-day resolutions.

Implemented and tested with OpenAI Codex.

### DIFF
--- a/src/akkudoktoreos/prediction/elecpriceenergycharts.py
+++ b/src/akkudoktoreos/prediction/elecpriceenergycharts.py
@@ -270,12 +270,22 @@ class ElecPriceEnergyCharts(ElecPriceProvider):
                 elif force_update:
                     # Use default start date in case of forced update
                     needs_update = True
-                elif not self._has_complete_published_horizon(
-                    now=now, resolution_seconds=resolution_seconds
-                ):
-                    # We have enough history, but not every expected source interval.
-                    start_datetime = gross_start_datetime
-                    needs_update = True
+                else:
+                    # The latest source data may have a different resolution than
+                    # the history before ems_start_datetime. Use its final 24 hours
+                    # so older, finer intervals cannot dominate the median, and
+                    # exclude the predicted tail.
+                    source_series = await self.key_to_raw_series(
+                        key="elecprice_marketprice_raw_wh",
+                        start_datetime=to_datetime(self.highest_orig_datetime).subtract(hours=24),
+                        end_datetime=to_datetime(self.highest_orig_datetime).add(seconds=1),
+                    )
+                    source_resolution_seconds = self._resolution_seconds(source_series)
+                    if not self._has_complete_published_horizon(
+                        now=now, resolution_seconds=source_resolution_seconds
+                    ):
+                        start_datetime = gross_start_datetime
+                        needs_update = True
         else:
             needs_update = True
 
--- a/tests/test_elecpriceenergycharts.py
+++ b/tests/test_elecpriceenergycharts.py
@@ -159,6 +159,7 @@ class TestElecPriceEnergyCharts:
 
     @pytest.mark.asyncio
     @pytest.mark.parametrize("host_timezone", ["UTC", "Europe/Berlin"])
+    @pytest.mark.parametrize("history_interval_minutes", [15, 60])
     @pytest.mark.parametrize(
         ("now", "last_price", "interval_minutes", "needs_update"),
         [
@@ -180,6 +181,7 @@ class TestElecPriceEnergyCharts:
         provider: ElecPriceEnergyCharts,
         set_other_timezone: Callable[[str], str],
         host_timezone: str,
+        history_interval_minutes: int,
         now: str,
         last_price: str,
         interval_minutes: int,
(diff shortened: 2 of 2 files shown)


dr-dimitri:
Thanks — this fixes the full-day resolution transition described in the original review comment and correctly excludes the predicted tail.

While this PR was being prepared, we independently created #1279 for the same underlying problem. That issue also tracks the short, within-day transition that this PR's description explicitly leaves open. This is an existing edge case, not a regression introduced by this PR, but #1279 should not be considered fully resolved by the current implementation.

### Remaining issue: the final 24-hour median can still use the old cadence

At `5e071e2`, `_update_data()` takes the final 24 hours of original source data and passes them to the shared median-based `_resolution_seconds()`. If a short, consistent new cadence follows many old-cadence intervals within that window, the old cadence still dominates.

### Deterministic reproduction

Use the existing `ElecPriceEnergyCharts` test fixture with an isolated database/configuration:

1. Set the provider timezone to `Europe/Berlin`, the EMS start to `2026-01-15 00:00:00+01:00`, and freeze the provider's clock at `2026-01-15 10:00:00+01:00`. Call `_update_data(force_update=False)`.
2. Seed `elecprice_marketprice_raw_wh` with hourly originals from `2025-12-11 00:00+01:00` through `2026-01-15 21:00+01:00`, inclusive. This supplies sufficient pre-EMS history, so the history-repair branch does not hide the coverage decision.
3. Append exactly five quarter-hourly original timestamps: `2026-01-15 22:00, 22:15, 22:30, 22:45, 23:00`, all in `Europe/Berlin`. Constant raw prices are sufficient. Set `highest_orig_datetime` to the final timestamp.
4. Mock `_request_forecast` and `_predict`; do not contact the live API. For a full update-path test, let the mocked API return quarter-hourly prices through `23:45`, using the real parsing/storage path.

**Actual:** the 24-hour median is 3,600 seconds. Coverage is computed as `23:00 + 1 hour = 00:00`, so no refresh is requested.

**Expected:** the latest original cadence is 900 seconds. Coverage ends at `23:15`, so a refresh is required to obtain the missing `23:15, 23:30, 23:45` intervals.

Repeat with the clock at exactly `2026-01-15 14:00+01:00` and move the transition/final timestamps to January 16, extending the old hourly series up to that transition. The same error affects the next-day publication horizon.

The reverse transition also reproduces: retain quarter-hourly originals through `18:45`, then append hourly originals at `19:00, 20:00, 21:00, 22:00, 23:00`. The median remains 900 seconds and triggers an unnecessary refresh, even though the latest hourly interval covers through midnight.

### Suggested fix and regression coverage

- Infer **coverage cadence** from a consistent trailing run of original intervals, rather than requiring the new cadence to dominate the whole 24-hour sample. For example, two matching final positive spacings could establish a supported new cadence; explicitly define a fallback for insufficient or ambiguous evidence. A single exceptional gap must not redefine an otherwise stable cadence.
- Keep the source bound at `highest_orig_datetime` and exclude predicted values after it. Add an opposing-cadence predicted tail to the tests so this remains enforced.
- Test five-point hourly-to-quarter-hourly transitions ending at both `23:00` (fetch) and `23:45` (no fetch), plus the reverse transition ending at `23:00` (no fetch), before and at the 14:00 cutoff. Keep the old-cadence data immediately before the new run; do not leave a large gap that accidentally removes it from the 24-hour sample.
- Keep history sufficiency, the shared forecasting/resampling resolution policy, forced updates, fetch-failure fallback, and existing timezone/DST behavior unchanged.
- #1279 also asks to reuse an unchanged source series where practical and reload it after a successful fetch; the current PR still performs a separate source read for coverage.

Validation note: this LLM-assisted review executed the unchanged PR decision methods from `5e071e2` in an isolated harness. The missed-refresh and unnecessary-refresh cases reproduced both before and at 14:00. This was a focused decision-path check, not a rerun of the full pytest suite.

PR #1283, opened 2026-09-06
fix: unify mypy environments for local checks and CI

### LINKED ISSUE #1276: [BUG]: make mypy and the pre-commit hook use inconsistent typing environments

dr-dimitri:
The isolated pre-commit mypy hook previously omitted runtime type information that `make mypy` used, hiding errors involving dependencies such as Pydantic and Pendulum. Makefile, pre-commit and CI now run the same full-project typing policy in the development environment defined by `uv.lock`.

- Use `uv run --locked --exact --extra dev` and the same mypy arguments for Makefile and the local hook. Check all of `src` and `tests`, including on configuration-only changes.
- Pin Python 3.13.15 for local development and the pre-commit CI job, and install the locked pre-commit version in CI.
- Disable incremental analysis because existing Pendulum cache state changes mypy 2.3.1 diagnostics. Document the policy, the performance tradeoff and the existing typing debt.
- Add a regression test that exercises Makefile, the hook and the CI command in a temporary project, accepting valid dependency types and detecting deliberate Pydantic/Pendulum assignment errors.

The full analysis exposes **956 existing errors in 125 files** on the validation checkout. Both typing commands and the pre-commit CI job therefore fail until that debt is addressed separately. The historical 961-error result recorded in the issue differs because of cached Pendulum timezone diagnostics; no additional checks or imports are suppressed.

Merge order:

1. Implement the existing typing-debt cleanup on the separate `fix/typing-debt` branch, starting from `main`, and validate it using the locked environment and full-analysis policy from this PR.
2. Review and merge the typing cleanup into `main` first, using separate, focused cleanup PRs where useful.
3. Update this PR from the resulting `main` and rerun Makefile, pre-commit and CI. The mypy check must pass before merging the configuration change.

This PR remains limited to the typing environment and policy fix for #1276. The existing type errors will be corrected separately rather than suppressed or folded into this PR.

Validation:

- `uv run --locked --extra dev pytest -q --finalize tests/test_mypy_toolchain.py` passes, including validation with an empty package cache.
- Makefile and pre-commit report equivalent full-project diagnostics with existing and fresh environments/caches; the cleaned PR branch also reports the same 956 errors from both entry points.
- Formatting and lint hooks for the changed files pass. Mypy fails only with the existing debt described above.
- Sphinx documentation build passes.

Closes #1276.

This pull request was created by an LLM (OpenAI Codex).


b0661:
Thanks you for the PR. It really reveals a lot of error spots and several false positives. We can not merge this in main currently as we would have unrelated CI fails on newly generated code.

Is there a way to keep CI passing for the current code but have an extra test that reveals the error spots. The fail of the extra test may be acceptable as long as not all the fixes are done but is a constant reminder to do the fixes.

dr-dimitri:
The prerequisite typing cleanup is now available in #1284: https://github.com/Akkudoktor-EOS/EOS/pull/1284.

This PR (#1283, implementing #1276) was blocked by its failing mypy CI check because the unified full development environment exposed 956 existing diagnostics. #1284 resolves that typing debt separately, without folding the runtime/model cleanup into this tooling-policy PR.

Please review and merge #1284 into `main` first, then update this PR from `main` and rerun its checks. This PR should merge only after the shared Makefile/pre-commit/CI typing policy passes.

The cleanup passes full, non-incremental mypy for 220 files and the existing hooks locally. Its description includes the runtime changes, explicit typing boundaries and validation limitations: local macOS server tests encounter a `psutil.AccessDenied` issue also reproduced on unchanged `main`, so Linux CI still needs to confirm the cleanup.

PR #1284, opened 2026-09-07
fix: resolve typing debt exposed by full-environment mypy

dr-dimitri:
The full development environment used by #1283 (the implementation of #1276) exposes 956 existing mypy diagnostics that were hidden by the isolated pre-commit environment. This separate cleanup resolves those diagnostics so the environment-unification PR can proceed after this change is merged into `main`.

#1283 is currently blocked by its failing mypy CI check. It remains focused on the tooling policy; this PR contains the prerequisite typing cleanup, based on `main`.

Changes:

- Enable the numpydantic and Pydantic mypy plugins, retaining strict Pydantic constructor typing with `init_typed = true`. Validate raw/coercible payloads through `model_validate`.
- Propagate concrete record, provider and time-window types through generic collections, factories and lookup methods. Preserve runtime field inspection and generated time-window documentation.
- Align Pendulum annotations with actual factory/arithmetic results while retaining Pydantic validation adapters at runtime. Correct optional values, array boundaries, REST handlers and plotting interfaces.
- Add pinned `scipy-stubs` and `types-psutil`, update `uv.lock`, and supply the plugins' dependencies to the existing isolated hook so this cleanup is checkable before #1283 changes that hook.
- Add runtime regression coverage for validated path defaults, normalized time-series metadata, generic field inspection, invalid timestamps and unsupported provider imports.

Runtime and compatibility details:

- Validate path defaults as `Path` objects while retaining raw string defaults needed by migration serialization with `exclude_defaults`.
- Normalize feed-in tariff lists and default charge rates to NumPy arrays; reject missing timestamps/uninitialized values explicitly. Importing into a provider without import support returns HTTP 400.
- Public JSON schemas and OpenAPI structure match `main` (excluding the generated version). All 20 regenerated configuration documents match regeneration of `main` in the same environment. No generated version/documentation churn or changes for #1277 are included.

Typing boundaries remain explicit: missing typing information is ignored only for the additional dependency groups `cachebox`, `fasthtml`, `monsterui`, `pvlib` and `statsmodels`. Five commented `override` exceptions preserve four existing collection iterator contracts and callable Pydantic Settings sources; narrowly scoped casts describe dynamic dependency boundaries. The full `src`/`tests` scope remains enabled. Passing mypy does not mean every third-party boundary is statically checked.

Validation on Python 3.13.15 / macOS:

- `uv run --locked --extra dev --exact --python 3.13.15 python -m mypy --no-incremental --python-version 3.13 --config-file pyproject.toml`: **passes, 220 files**.
- `make mypy` and `make format`, including the existing pre-commit mypy hook: **pass**. The new test file also passes the hooks explicitly.
- Final targeted run with the `make test-ci` options for configuration, migration, Pydantic and the new/documentation tests: **109 passed, 3 skipped**.
- Full `make test-ci` run: **1,561 passed, 18 skipped, 5 failed, 19 setup errors**. Four failures identified during that run were corrected and pass in the final targeted run. The remaining failure and all 19 setup errors share macOS `psutil.AccessDenied` from system-wide connection inspection; a representative server-start test reproduces it on unchanged `main`.
- Sphinx build and docstring validation pass. Eight rendered chart comparisons against `main` are pixel-identical, covering missing values and a DST transition in both optimization variants. The two PDF comparison tests skip locally because their converter is unavailable.
- `GITHUB_ACTIONS=true` reproduces the three existing generated-document freshness skips from CI. A completely green full-suite run is **not** claimed locally; Linux CI must validate this PR before merge.

Merge order:

1. Review this typing cleanup and merge it into `main` after CI validation.
2. Update #1283 from the resulting `main` and rerun its checks.
3. Merge #1283 once its shared Makefile/pre-commit/CI typing policy passes.

Related: #1283 and #1276. This PR does not replace or close the tooling-policy issue.

This pull request was created by an LLM (OpenAI Codex).

### DIFF
--- a/scripts/generate_config_md.py
+++ b/scripts/generate_config_md.py
@@ -467,7 +467,8 @@ def write_to_file(file_path: Optional[Union[str, Path]], config_md: str):
 
     # Assure timezone name does not leak to documentation
     tz_name = to_datetime().timezone_name
-    assert tz_name is not None
+    if tz_name is None:
+        raise RuntimeError("Documentation generation requires a timezone name")
     config_md = re.sub(re.escape(tz_name), "Europe/Berlin", config_md, flags=re.IGNORECASE)
     # Also replace UTC, as GitHub CI always is on UTC
     config_md = re.sub(re.escape("UTC"), "Europe/Berlin", config_md, flags=re.IGNORECASE)
--- a/scripts/generate_config_md.py
+++ b/scripts/generate_config_md.py
@@ -8,7 +8,7 @@ import re
 import sys
 import textwrap
 from pathlib import Path
-from typing import Any, Optional, Type, Union, get_args
+from typing import Any, Optional, Type, TypeVar, Union, get_args
 
 from loguru import logger
 from pydantic.fields import ComputedFieldInfo, FieldInfo
@@ -24,8 +24,8 @@ from akkudoktoreos.core.coreabc import get_config, singletons_init
 from akkudoktoreos.core.pydantic import PydanticBaseModel
 from akkudoktoreos.utils.datetimeutil import to_datetime
 
-documented_types: set[PydanticBaseModel] = set()
-undocumented_types: dict[PydanticBaseModel, tuple[str, list[str]]] = dict()
+documented_types: set[type[PydanticBaseModel]] = set()
+undocumented_types: dict[type[PydanticBaseModel], tuple[str, list[str]]] = dict()
 
 global_config_dict: dict[str, Any] = dict()
 
@@ -60,6 +60,9 @@ def get_body(config: type[PydanticBaseModel]) -> str:
 
 def resolve_nested_types(field_type: Any, parent_types: list[str]) -> list[tuple[Any, list[str]]]:
     resolved_types: list[tuple[type, list[str]]] = []
+
+    if isinstance(field_type, TypeVar):
+        field_type = field_type.__bound__ or Any
 
     origin = getattr(field_type, "__origin__", field_type)
     if origin is Union:
@@ -167,7 +170,7 @@ def build_nested_structure(keys: list[str], value: Any) -> Any:
 
 def get_default_value(field_info: Union[FieldInfo, ComputedFieldInfo], regular_field: bool) -> Any:
     default_value = ""
-    if regular_field:
+    if regular_field and isinstance(field_info, FieldInfo):
         if (val := field_info.default) is not PydanticUndefined:
             default_value = val
         else:
@@ -177,8 +180,13 @@ def get_default_value(field_info: Union[FieldInfo, ComputedFieldInfo], regular_f
     return default_value
 
 
-def get_type_name(field_type: type) -> str:
+def get_type_name(field_type: Any) -> str:
     type_name = str(field_type).replace("typing.", "").replace("pathlib._local", "pathlib")
+    # Unparameterized Pydantic generics validate against their TypeVar bound.
+    for arg in get_args(field_type):
+        if isinstance(arg, TypeVar) and isinstance(arg.__bound__, type):
+            bound = arg.__bound__
+            type_name = type_name.replace(str(arg), f"{bound.__module__}.{bound.__qualname__}")
     if type_name.startswith("<class"):
         type_name = field_type.__name__
     return type_name
@@ -229,17 +237,14 @@ def generate_config_table_md(
         table += f"| ---- {env_header_underline}| ---- | --------- | ------- | ----------- |\n"
 
 
-    fields = {}
-    for field_name, field_info in config.model_fields.items():
-        fields[field_name] = field_info
-    for field_name, field_info in config.model_computed_fields.items():
-        fields[field_name] = field_info
+    fields: dict[str, FieldInfo | ComputedFieldInfo] = dict(config.model_fields)
+    fields.update(config.model_computed_fields)
     for field_name in sorted(fields.keys()):
         field_info = fields[field_name]
         regular_field = isinstance(field_info, FieldInfo)
 
         config_name = field_name if extra_config else field_name.upper()
-        field_type = field_info.annotation if regular_field else field_info.return_type
+        field_type = (field_info.annotation if isinstance(field_info, FieldInfo) else field_info.return_type)
         default_value = get_default_value(field_info, regular_field)
         description = config.field_description(field_name)
         deprecated = config.field_deprecated(field_name)
@@ -462,6 +467,7 @@ def write_to_file(file_path: Optional[Union[str, Path]], config_md: str):
 
     # Assure timezone name does not leak to documentation
     tz_name = to_datetime().timezone_name
+    assert tz_name is not None
     config_md = re.sub(re.escape(tz_name), "Europe/Berlin", config_md, flags=re.IGNORECASE)
     # Also replace UTC, as GitHub CI always is on UTC
     config_md = re.sub(re.escape("UTC"), "Europe/Berlin", config_md, flags=re.IGNORECASE)
--- a/src/akkudoktoreos/adapter/homeassistant.py
+++ b/src/akkudoktoreos/adapter/homeassistant.py
@@ -145,7 +145,10 @@ class HomeAssistantAdapterCommonSettings(SettingsBaseModel):
         """Entity IDs available at Home Assistant."""
         try:
             adapter_eos = get_adapter()
-            result = adapter_eos.provider_by_id("HomeAssistant").get_homeassistant_entity_ids()
+            provider = adapter_eos.provider_by_id("HomeAssistant")
+            if not isinstance(provider, HomeAssistantAdapter):
+                return []
+            result = provider.get_homeassistant_entity_ids()
         except Exception:
             return []
         return result
@@ -156,7 +159,10 @@ class HomeAssistantAdapterCommonSettings(SettingsBaseModel):
         """Entity IDs for optimization solution available at EOS."""
         try:
             adapter_eos = get_adapter()
-            result = adapter_eos.provider_by_id("HomeAssistant").get_eos_solution_entity_ids()
+            provider = adapter_eos.provider_by_id("HomeAssistant")
+            if not isinstance(provider, HomeAssistantAdapter):
+                return []
+            result = provider.get_eos_solution_entity_ids()
         except Exception:
             return []
         return result
@@ -167,9 +173,10 @@ class HomeAssistantAdapterCommonSettings(SettingsBaseModel):
         """Entity IDs for energy management instructions available at EOS."""
         try:
             adapter_eos = get_adapter()
-            result = adapter_eos.provider_by_id(
-                "HomeAssistant"
-            ).get_eos_device_instruction_entity_ids()
+            provider = adapter_eos.provider_by_id("HomeAssistant")
+            if not isinstance(provider, HomeAssistantAdapter):
+                return []
+            result = provider.get_eos_device_instruction_entity_ids()
         except Exception:
             return []
         return result
--- a/src/akkudoktoreos/adapter/homeassistant.py
+++ b/src/akkudoktoreos/adapter/homeassistant.py
@@ -146,8 +146,11 @@ class HomeAssistantAdapterCommonSettings(SettingsBaseModel):
         try:
             adapter_eos = get_adapter()
             provider = adapter_eos.provider_by_id("HomeAssistant")
-            if not isinstance(provider, HomeAssistantAdapter):
-                return []
+        except Exception:
+            return []
+        if not isinstance(provider, HomeAssistantAdapter):
+            raise TypeError("HomeAssistant provider must be a HomeAssistantAdapter")
+        try:
             result = provider.get_homeassistant_entity_ids()
         except Exception:
             return []
@@ -160,8 +163,11 @@ class HomeAssistantAdapterCommonSettings(SettingsBaseModel):
         try:
             adapter_eos = get_adapter()
             provider = adapter_eos.provider_by_id("HomeAssistant")
-            if not isinstance(provider, HomeAssistantAdapter):
-                return []
+        except Exception:
+            return []
+        if not isinstance(provider, HomeAssistantAdapter):
+            raise TypeError("HomeAssistant provider must be a HomeAssistantAdapter")
+        try:
             result = provider.get_eos_solution_entity_ids()
         except Exception:
             return []
@@ -174,8 +180,11 @@ class HomeAssistantAdapterCommonSettings(SettingsBaseModel):
         try:
             adapter_eos = get_adapter()
             provider = adapter_eos.provider_by_id("HomeAssistant")
-            if not isinstance(provider, HomeAssistantAdapter):
-                return []
+        except Exception:
+            return []
+        if not isinstance(provider, HomeAssistantAdapter):
+            raise TypeError("HomeAssistant provider must be a HomeAssistantAdapter")
+        try:
             result = provider.get_eos_device_instruction_entity_ids()
         except Exception:
             return []
--- a/src/akkudoktoreos/config/config.py
+++ b/src/akkudoktoreos/config/config.py
@@ -14,7 +14,7 @@ import os
 import sys
 import tempfile
 from pathlib import Path
-from typing import Any, ClassVar, Optional, Type, Union
+from typing import Any, Callable, ClassVar, Optional, Type, Union
 
 import pydantic_settings
 from loguru import logger
@@ -122,6 +122,10 @@ def default_data_folder_path() -> Path:
 class GeneralSettings(SettingsBaseModel):
     """General settings."""
 
+    # Legacy configuration-path metadata populated by ConfigEOS._setup_config_file.
+    _config_file_path: ClassVar[Path | None] = None
+    _config_folder_path: ClassVar[Path | None] = None
+
     config_save_mode: ConfigSaveMode = Field(
         default=ConfigSaveMode.AUTOMATIC,
         json_schema_extra={
@@ -161,8 +165,11 @@ class GeneralSettings(SettingsBaseModel):
         },
     )
 
+    # Validate this raw default to Path. Retain the string so
+    # exclude_defaults preserves the output path in migrated configurations.
     data_output_subpath: Optional[Path] = Field(
         default="output",
+        validate_default=True,
         json_schema_extra={"description": "Sub-path for the EOS output data folder."},
     )
 
@@ -402,14 +409,15 @@ class ConfigEOS(SingletonMixin, SettingsEOSDefaults):
         return True
 
     @classmethod
-    def settings_customise_sources(
+    # Pydantic Settings accepts zero-argument callables as well as source objects.
+    def settings_customise_sources(  # type: ignore[override]
         cls,
         settings_cls: Type[pydantic_settings.BaseSettings],
         init_settings: pydantic_settings.PydanticBaseSettingsSource,
         env_settings: pydantic_settings.PydanticBaseSettingsSource,
         dotenv_settings: pydantic_settings.PydanticBaseSettingsSource,
         file_secret_settings: pydantic_settings.PydanticBaseSettingsSource,
-    ) -> tuple[pydantic_settings.PydanticBaseSettingsSource, ...]:
+    ) -> tuple[pydantic_settings.PydanticBaseSettingsSource | Callable[[], dict[str, Any]], ...]:
         """Customizes the order and handling of settings sources for a pydantic_settings.BaseSettings subclass.
 
         This method determines the sources for application configuration settings, including
@@ -790,7 +798,10 @@ class ConfigEOS(SingletonMixin, SettingsEOSDefaults):
                 required by ``self._setup()``.
             OSError: If reading the backup file fails due to I/O issues.
         """
-        backup_file_path = self.general.config_file_path.with_suffix(f".{backup_id}")
+        config_file_path = self.general.config_file_path
+        if config_file_path is None:
+            raise RuntimeError("Configuration file path is not initialized")
+        backup_file_path = config_file_path.with_suffix(f".{backup_id}")
         if not backup_file_path.exists():
             error_msg = f"Configuration backup `{backup_id}` not found."
             logger.error(error_msg)
@@ -823,7 +834,9 @@ class ConfigEOS(SingletonMixin, SettingsEOSDefaults):
         """
         result: dict[str, dict[str, Any]] = {}
 
-        base_path: Path = self.general.config_file_path
+        base_path = self.general.config_file_path
+        if base_path is None:
+            raise RuntimeError("Configuration file path is not initialized")
         parent = base_path.parent
         stem = base_path.stem
 
--- a/src/akkudoktoreos/config/config.py
+++ b/src/akkudoktoreos/config/config.py
@@ -799,8 +799,9 @@ class ConfigEOS(SingletonMixin, SettingsEOSDefaults):
             OSError: If reading the backup file fails due to I/O issues.
         """
         config_file_path = self.general.config_file_path
+        # Configuration setup initializes this path; should never raise.
         if config_file_path is None:
-            raise RuntimeError("Configuration file path is not initialized")
+            raise AssertionError("Configuration file path is not initialized")
         backup_file_path = config_file_path.with_suffix(f".{backup_id}")
         if not backup_file_path.exists():
             error_msg = f"Configuration backup `{backup_id}` not found."
@@ -835,8 +836,9 @@ class ConfigEOS(SingletonMixin, SettingsEOSDefaults):
         result: dict[str, dict[str, Any]] = {}
 
         base_path = self.general.config_file_path
+        # Configuration setup initializes this path; should never raise.
         if base_path is None:
-            raise RuntimeError("Configuration file path is not initialized")
+            raise AssertionError("Configuration file path is not initialized")
         parent = base_path.parent
         stem = base_path.stem
 
--- a/src/akkudoktoreos/config/configabc.py
+++ b/src/akkudoktoreos/config/configabc.py
@@ -409,23 +409,23 @@ class TimeWindow(SettingsBaseModel):
         return self.duration
 
 
-WindowT = TypeVar("WindowT", bound=TimeWindow)
+TimeWindowT = TypeVar("TimeWindowT", bound=TimeWindow)
 
 
-class TimeWindowSequence(SettingsBaseModel, Generic[WindowT]):
+class TimeWindowSequence(SettingsBaseModel, Generic[TimeWindowT]):
     """Model representing a sequence of time windows with collective operations.
 
     Manages multiple TimeWindow objects and provides methods to work with them
     as a cohesive unit for scheduling and availability checking.
     """
 
-    windows: list[WindowT] = Field(
+    windows: list[TimeWindowT] = Field(
         default_factory=list,
         json_schema_extra={"description": "List of TimeWindow objects that make up this sequence."},
     )
 
     # EOS collections iterate over their elements instead of BaseModel field/value pairs.
-    def __iter__(self) -> Iterator[WindowT]:  # type: ignore[override]
+    def __iter__(self) -> Iterator[TimeWindowT]:  # type: ignore[override]
         """Allow iteration over the time windows."""
         return iter(self.windows)
 
@@ -433,7 +433,7 @@ class TimeWindowSequence(SettingsBaseModel, Generic[WindowT]):
         """Return the number of time windows in the sequence."""
         return len(self.windows)
 
-    def __getitem__(self, index: int) -> WindowT:
+    def __getitem__(self, index: int) -> TimeWindowT:
         """Allow indexing into the time windows."""
         return self.windows[index]
 
@@ -540,7 +540,9 @@ class TimeWindowSequence(SettingsBaseModel, Generic[WindowT]):
             total += d
         return total
 
-    def get_applicable_windows(self, reference_date: Optional[DateTime] = None) -> list[WindowT]:
+    def get_applicable_windows(
+        self, reference_date: Optional[DateTime] = None
+    ) -> list[TimeWindowT]:
         """Get all windows that apply to the given reference date.
 
         Args:
@@ -560,7 +562,7 @@ class TimeWindowSequence(SettingsBaseModel, Generic[WindowT]):
 
     def find_windows_for_duration(
         self, duration: Duration, reference_date: Optional[DateTime] = None
-    ) -> list[WindowT]:
+    ) -> list[TimeWindowT]:
         """Find all windows that can accommodate the given duration.
 
         Args:
@@ -579,7 +581,7 @@ class TimeWindowSequence(SettingsBaseModel, Generic[WindowT]):
 
     def get_all_possible_start_times(
         self, duration: Duration, reference_date: Optional[DateTime] = None
-    ) -> list[tuple[DateTime, DateTime, WindowT]]:
+    ) -> list[tuple[DateTime, DateTime, TimeWindowT]]:
         """Get all possible start time ranges for a duration across all windows.
 
         Args:
@@ -743,7 +745,7 @@ class TimeWindowSequence(SettingsBaseModel, Generic[WindowT]):
             dtype=np.float64,
         )
 
-    def add_window(self, window: WindowT) -> None:
+    def add_window(self, window: TimeWindowT) -> None:
         """Add a new time window to the sequence.
 
         Args:
@@ -751,7 +753,7 @@ class TimeWindowSequence(SettingsBaseModel, Generic[WindowT]):
         """
         self.windows.append(window)
 
-    def remove_window(self, index: int) -> WindowT:
+    def remove_window(self, index: int) -> TimeWindowT:
         """Remove a time window from the sequence by index.
 
         Args:
@@ -785,7 +787,7 @@ class TimeWindowSequence(SettingsBaseModel, Generic[WindowT]):
         if reference_date is None:
             reference_date = pendulum.today()
 
-        def sort_key(window: WindowT) -> tuple[int, DateTime]:
+        def sort_key(window: TimeWindowT) -> tuple[int, DateTime]:
             start_time = window.earliest_start_time(Duration(), reference_date)
             if start_time is None:
                 return (1, reference_date)
--- a/src/akkudoktoreos/config/configabc.py
+++ b/src/akkudoktoreos/config/configabc.py
@@ -4,7 +4,7 @@ import calendar
 import os
 import sys
 from enum import StrEnum
-from typing import Any, ClassVar, Iterator, Optional, Union
+from typing import Any, ClassVar, Generic, Iterator, Optional, TypeVar, Union
 
 import numpy as np
 import pandas as pd
@@ -409,19 +409,23 @@ class TimeWindow(SettingsBaseModel):
         return self.duration
 
 
-class TimeWindowSequence(SettingsBaseModel):
+WindowT = TypeVar("WindowT", bound=TimeWindow)
+
+
+class TimeWindowSequence(SettingsBaseModel, Generic[WindowT]):
     """Model representing a sequence of time windows with collective operations.
 
     Manages multiple TimeWindow objects and provides methods to work with them
     as a cohesive unit for scheduling and availability checking.
     """
 
-    windows: list[TimeWindow] = Field(
+    windows: list[WindowT] = Field(
         default_factory=list,
         json_schema_extra={"description": "List of TimeWindow objects that make up this sequence."},
     )
 
-    def __iter__(self) -> Iterator[TimeWindow]:
+    # EOS collections iterate over their elements instead of BaseModel field/value pairs.
+    def __iter__(self) -> Iterator[WindowT]:  # type: ignore[override]
         """Allow iteration over the time windows."""
         return iter(self.windows)
 
@@ -429,7 +433,7 @@ class TimeWindowSequence(SettingsBaseModel):
         """Return the number of time windows in the sequence."""
         return len(self.windows)
 
-    def __getitem__(self, index: int) -> TimeWindow:
+    def __getitem__(self, index: int) -> WindowT:
         """Allow indexing into the time windows."""
         return self.windows[index]
 
@@ -536,7 +540,7 @@ class TimeWindowSequence(SettingsBaseModel):
             total += d
         return total
 
-    def get_applicable_windows(self, reference_date: Optional[DateTime] = None) -> list[TimeWindow]:
+    def get_applicable_windows(self, reference_date: Optional[DateTime] = None) -> list[WindowT]:
         """Get all windows that apply to the given reference date.
 
         Args:
@@ -556,7 +560,7 @@ class TimeWindowSequence(SettingsBaseModel):
 
     def find_windows_for_duration(
         self, duration: Duration, reference_date: Optional[DateTime] = None
-    ) -> list[TimeWindow]:
+    ) -> list[WindowT]:
         """Find all windows that can accommodate the given duration.
 
         Args:
@@ -575,7 +579,7 @@ class TimeWindowSequence(SettingsBaseModel):
 
     def get_all_possible_start_times(
         self, duration: Duration, reference_date: Optional[DateTime] = None
-    ) -> list[tuple[DateTime, DateTime, TimeWindow]]:
+    ) -> list[tuple[DateTime, DateTime, WindowT]]:
         """Get all possible start time ranges for a duration across all windows.
 
         Args:
@@ -739,7 +743,7 @@ class TimeWindowSequence(SettingsBaseModel):
             dtype=np.float64,
         )
 
-    def add_window(self, window: TimeWindow) -> None:
+    def add_window(self, window: WindowT) -> None:
         """Add a new time window to the sequence.
 
         Args:
@@ -747,7 +751,7 @@ class TimeWindowSequence(SettingsBaseModel):
         """
         self.windows.append(window)
 
-    def remove_window(self, index: int) -> TimeWindow:
+    def remove_window(self, index: int) -> WindowT:
         """Remove a time window from the sequence by index.
 
         Args:
@@ -781,7 +785,7 @@ class TimeWindowSequence(SettingsBaseModel):
         if reference_date is None:
             reference_date = pendulum.today()
 
-        def sort_key(window: TimeWindow) -> tuple[int, DateTime]:
+        def sort_key(window: WindowT) -> tuple[int, DateTime]:
             start_time = window.earliest_start_time(Duration(), reference_date)
             if start_time is None:
                 return (1, reference_date)
@@ -806,7 +810,7 @@ class ValueTimeWindow(TimeWindow):
     )
 
 
-class ValueTimeWindowSequence(TimeWindowSequence):
+class ValueTimeWindowSequence(TimeWindowSequence[ValueTimeWindow]):
     """Sequence of value time windows.
 
     This model specializes `TimeWindowSequence` to ensure that all
--- a/src/akkudoktoreos/core/cache.py
+++ b/src/akkudoktoreos/core/cache.py
@@ -742,6 +742,8 @@ class CacheFileStore(ConfigMixin, SingletonMixin):
                 if clear_all:
                     clear_file = True
                 else:
+                    if before_datetime is None:
+                        raise RuntimeError("Cache expiry threshold is not initialized")
                     clear_file = compare_datetimes(cache_item.until_datetime, before_datetime).lt
 
                 if clear_file:
@@ -782,9 +784,11 @@ class CacheFileStore(ConfigMixin, SingletonMixin):
         with self._store_lock:
             store_current = {}
             for key, record in self._store.items():
-                ttl_duration = record.ttl_duration
-                if ttl_duration:
-                    ttl_duration = ttl_duration.total_seconds()
+                ttl_duration = (
+                    record.ttl_duration.total_seconds()
+                    if record.ttl_duration
+                    else record.ttl_duration
+                )
                 store_current[key] = {
                     # Convert file-like objects to file paths for serialization
                     "cache_file": self._get_file_path(record.cache_file),
--- a/src/akkudoktoreos/core/cache.py
+++ b/src/akkudoktoreos/core/cache.py
@@ -742,8 +742,9 @@ class CacheFileStore(ConfigMixin, SingletonMixin):
                 if clear_all:
                     clear_file = True
                 else:
+                    # Initialized above when clear_all is false; should never raise.
                     if before_datetime is None:
-                        raise RuntimeError("Cache expiry threshold is not initialized")
+                        raise AssertionError("Cache expiry threshold is not initialized")
                     clear_file = compare_datetimes(cache_item.until_datetime, before_datetime).lt
 
                 if clear_file:
--- a/src/akkudoktoreos/core/dataabc.py
+++ b/src/akkudoktoreos/core/dataabc.py
@@ -20,10 +20,12 @@ from typing import (
     TYPE_CHECKING,
     Any,
     Dict,
+    Generic,
     Iterator,
     Optional,
     Tuple,
     Type,
+    TypeVar,
     Union,
     get_args,
     overload,
@@ -289,7 +291,8 @@ class DataRecord(DataABC, MutableMapping):
         except AttributeError:
             raise KeyError(f"'{key}' is not a recognized field.")
 
-    def __iter__(self) -> Iterator[str]:
+    # EOS collections iterate over their elements instead of BaseModel field/value pairs.
+    def __iter__(self) -> Iterator[str]:  # type: ignore[override]
         """Iterate over the field names in the data record.
 
         Returns:
@@ -441,7 +444,10 @@ class DataRecord(DataABC, MutableMapping):
 # ==================== DataSequence ====================
 
 
-class DataSequence(DataABC, DatabaseRecordProtocolMixin[DataRecord]):
+RecordT = TypeVar("RecordT", bound=DataRecord)
+
+
+class DataSequence(DataABC, DatabaseRecordProtocolMixin[RecordT], Generic[RecordT]):
     """A managed sequence of DataRecord instances with time series behavior.
 
     The DataSequence class provides an ordered, mutable collection of DataRecord
@@ -488,7 +494,7 @@ class DataSequence(DataABC, DatabaseRecordProtocolMixin[DataRecord]):
     """
 
     # To be overloaded by derived classes.
-    records: list[DataRecord] = Field(
+    records: list[RecordT] = Field(
         default_factory=list, json_schema_extra={"description": "List of data records"}
     )
 
@@ -553,7 +559,7 @@ class DataSequence(DataABC, DatabaseRecordProtocolMixin[DataRecord]):
                 f"Key '{key}' is not in writable record keys: {self.record_keys_writable}"
             )
 
-    def _validate_record(self, value: DataRecord) -> None:
+    def _validate_record(self, value: RecordT) -> None:
         """Check if the provided value is a valid DataRecord with compatible keys.
 
         Args:
@@ -629,7 +635,7 @@ class DataSequence(DataABC, DatabaseRecordProtocolMixin[DataRecord]):
         return self.record_class().record_keys_writable()
 
     @classmethod
-    def record_class(cls) -> Type:
+    def record_class(cls) -> Type[RecordT]:
         """Get the class of the data record handled by this data sequence.
 
         This method determines the class of the data record type associated with
@@ -646,6 +652,8 @@ class DataSequence(DataABC, DatabaseRecordProtocolMixin[DataRecord]):
         field_info = cls.model_fields["records"]
         # Get the list element type from the 'type_' attribute
         list_element_type = get_args(field_info.annotation)[0]
+        if isinstance(list_element_type, TypeVar):
+            list_element_type = list_element_type.__bound__
         if not isinstance(list_element_type(), DataRecord):
             raise ValueError(
                 f"Data record must be an instance of DataRecord: '{list_element_type}'."
@@ -736,17 +744,18 @@ class DataSequence(DataABC, DatabaseRecordProtocolMixin[DataRecord]):
 
     # Sequence methods
 
-    def __iter__(self) -> Iterator[DataRecord]:
+    # EOS collections iterate over their elements instead of BaseModel field/value pairs.
+    def __iter__(self) -> Iterator[RecordT]:  # type: ignore[override]
         """Create an iterator for accessing DataRecords sequentially (memory only).
 
         Returns:
-            Iterator[DataRecord]: An iterator for the records.
+            Iterator[RecordT]: An iterator for the records.
         """
         return iter(self.records)
 
     async def get_by_datetime(
         self, target_datetime: DateTime, *, time_window: Optional[Duration] = None
-    ) -> Optional[DataRecord]:
+    ) -> Optional[RecordT]:
         """Get the record at the specified datetime, with an optional fallback search window.
 
         Args:
@@ -770,7 +779,7 @@ class DataSequence(DataABC, DatabaseRecordProtocolMixin[DataRecord]):
 
     async def get_nearest_by_datetime(
         self, target_datetime: DateTime, time_window: Optional[Duration] = None
-    ) -> Optional[DataRecord]:
+    ) -> Optional[RecordT]:
         """Get the record nearest to the specified datetime within an optional time window.
 
         Args:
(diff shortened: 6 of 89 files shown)


src/akkudoktoreos/adapter/homeassistant.py
```diff
@@ -145,7 +145,10 @@ def homeassistant_entity_ids(self) -> list[str]:
         """Entity IDs available at Home Assistant."""
         try:
             adapter_eos = get_adapter()
-            result = adapter_eos.provider_by_id("HomeAssistant").get_homeassistant_entity_ids()
+            provider = adapter_eos.provider_by_id("HomeAssistant")
+            if not isinstance(provider, HomeAssistantAdapter):
+                return []
```
b0661: This shall raise an error. It is a clear implementation error.

src/akkudoktoreos/adapter/homeassistant.py
```diff
@@ -156,7 +159,10 @@ def eos_solution_entity_ids(self) -> list[str]:
         """Entity IDs for optimization solution available at EOS."""
         try:
             adapter_eos = get_adapter()
-            result = adapter_eos.provider_by_id("HomeAssistant").get_eos_solution_entity_ids()
+            provider = adapter_eos.provider_by_id("HomeAssistant")
+            if not isinstance(provider, HomeAssistantAdapter):
+                return []
```
b0661: This shall raise an error. It is a clear implementation error.

src/akkudoktoreos/adapter/homeassistant.py
```diff
@@ -167,9 +173,10 @@ def eos_device_instruction_entity_ids(self) -> list[str]:
         """Entity IDs for energy management instructions available at EOS."""
         try:
             adapter_eos = get_adapter()
-            result = adapter_eos.provider_by_id(
-                "HomeAssistant"
-            ).get_eos_device_instruction_entity_ids()
+            provider = adapter_eos.provider_by_id("HomeAssistant")
+            if not isinstance(provider, HomeAssistantAdapter):
+                return []
```
b0661: This shall raise an error. It is a clear implementation error.

src/akkudoktoreos/core/cache.py
```diff
@@ -742,6 +742,8 @@ def clear(
                 if clear_all:
                     clear_file = True
                 else:
+                    if before_datetime is None:
+                        raise RuntimeError("Cache expiry threshold is not initialized")
```
b0661: This can never happen, but mypy does not detect that. It would be a clear implementation error if this triggers. Don't use RuntimeError here. Please also add a comment e.g. "make mypy happy, should never raise"

src/akkudoktoreos/core/cache.py
```diff
@@ -782,9 +784,11 @@ def current_store(self) -> dict:
         with self._store_lock:
             store_current = {}
             for key, record in self._store.items():
-                ttl_duration = record.ttl_duration
-                if ttl_duration:
-                    ttl_duration = ttl_duration.total_seconds()
+                ttl_duration = (
```
b0661: This bails out on ttl_duration == None. Please do not change.

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -27,6 +27,7 @@
 
 from loguru import logger
 from numpydantic import NDArray, Shape
+from pendulum import UTC
```
b0661: No pendulum imports in regular modules. All imports have to be done by datetimeutil. If you absolutely need UTC instead of "UTC" make it available in datetimeutil.py.

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -302,16 +311,14 @@ class DatabaseTimestamp(str):
     __slots__ = ()
 
     @classmethod
-    def from_datetime(cls, dt: DateTime) -> "DatabaseTimestamp":
-        if dt.tz is None:
+    def from_datetime(cls, dt: DateTime | None) -> "DatabaseTimestamp":
```
b0661: Do not widen the type with None.  Better replace Any by a stricter definition for DateTime.

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -278,7 +279,7 @@ def deserialize_data(self, data: bytes) -> bytes:
 
 
 class DataRecordProtocol(Protocol):
-    date_time: DateTime
+    date_time: DateTime | None
```
b0661, line 284: Do not widen the type with None. We need a datetime for a data record. Better replace Any by a stricter definition for DateTime.

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -287,6 +288,14 @@ def __getitem__(self, key: str) -> Any: ...
     def model_dump(self) -> dict: ...
 
 
+def _record_datetime(record: DataRecordProtocol) -> DateTime:
```
b0661, line 293: This is a simple solution but shifts the type checking to runtime. Please see above.

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -1407,7 +1414,9 @@ async def db_get_record(
         record = min(
             candidates,
             key=lambda r: abs(
-                (r.date_time - DatabaseTimestamp.to_datetime(target_timestamp)).total_seconds()
+                (
```
b0661, line 1420: Same issue here. Do not widen the type with None. See above.

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -1417,7 +1426,7 @@ async def db_get_record(
             if (
                 abs(
                     (
-                        record.date_time - DatabaseTimestamp.to_datetime(target_timestamp)
+                        _record_datetime(record) - DatabaseTimestamp.to_datetime(target_timestamp)
```
b0661: Same issue here. Do not widen the type with None. See above.

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -1969,7 +1978,7 @@ async def _db_compact_tier(
         # run — they are inside the age window but straddle an incomplete bucket.
         raw_cutoff_epoch = int(raw_cutoff_dt.timestamp())
         floored_cutoff_epoch = (raw_cutoff_epoch // interval_sec) * interval_sec
-        new_cutoff_dt = DateTime.fromtimestamp(floored_cutoff_epoch, tz="UTC")
+        new_cutoff_dt = DateTime.fromtimestamp(floored_cutoff_epoch, tz=UTC)
```
b0661, line 1988: See above. No import from pendulum in regular modules.

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -2001,7 +2010,7 @@ async def _db_compact_tier(
         # overwritten with the same values).
         raw_start_epoch = int(raw_window_start_dt.timestamp())
         floored_start_epoch = (raw_start_epoch // interval_sec) * interval_sec
-        window_start_dt = DateTime.fromtimestamp(floored_start_epoch, tz="UTC")
+        window_start_dt = DateTime.fromtimestamp(floored_start_epoch, tz=UTC)
```
b0661, line 2020: See above. No import from pendulum in regular modules.

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -2065,8 +2076,8 @@ async def _db_compact_tier(
             # Process chronologically so the earliest record's values win when
             # multiple records floor to the same bucket.
             snapped_bucket: dict[int, dict[str, Any]] = {}
-            for r in sorted(records_in_window, key=lambda x: x.date_time):
-                ts_epoch = int(r.date_time.timestamp())
+            for r in sorted(records_in_window, key=_record_datetime):
+                ts_epoch = int(_record_datetime(r).timestamp())
```
b0661: Same issue here. Do not widen the type with None. See above.

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -2037,7 +2046,9 @@ async def _db_compact_tier(
                 if r.date_time is not None and window_start_dt <= r.date_time < window_end_dt
             ]
             misaligned = [
-                r for r in records_in_window if int(r.date_time.timestamp()) % interval_sec != 0
+                r
+                for r in records_in_window
+                if int(_record_datetime(r).timestamp()) % interval_sec != 0
```
b0661: Same issue here. Do not widen the type with None. See above.

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -2089,7 +2100,7 @@ async def _db_compact_tier(
             for snapped_epoch, values in snapped_bucket.items():
                 if not values:
                     continue
-                snapped_dt = DateTime.fromtimestamp(snapped_epoch, tz="UTC")
+                snapped_dt = DateTime.fromtimestamp(snapped_epoch, tz=UTC)
```
b0661, line 2111: See above. No import from pendulum in regular modules.

src/akkudoktoreos/core/databaseabc.py
```diff
@@ -2148,7 +2159,7 @@ async def _db_compact_tier(
                 while first_bucket_epoch < int(window_start_dt.timestamp()):
                     first_bucket_epoch += interval_sec
                 compacted_timestamps = [
-                    DateTime.fromtimestamp(first_bucket_epoch + i * interval_sec, tz="UTC")
+                    DateTime.fromtimestamp(first_bucket_epoch + i * interval_sec, tz=UTC)
```
b0661, line 2170: See above. No import from pendulum in regular modules.

src/akkudoktoreos/core/ems.py
```diff
@@ -93,6 +93,8 @@ def start_datetime(self) -> DateTime:
         """The starting datetime of the current or latest energy management."""
         if EnergyManagement._start_datetime is None:
             EnergyManagement.set_start_datetime()
+        if EnergyManagement._start_datetime is None:
```
b0661: This can never happen, but mypy does not detect that. It would be a clear implementation error if this triggers. Don't use RuntimeError here. Please also add a comment e.g. "make mypy happy, should never raise"

src/akkudoktoreos/prediction/predictionabc.py
```diff
@@ -21,6 +21,7 @@
     DataProvider,
     DataRecord,
     DataSequence,
+    RecordT,
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/prediction/predictionabc.py
```diff
@@ -185,7 +189,7 @@ def keep_hours(self) -> Optional[int]:
         return int(duration.total_hours())
 
 
-class PredictionProvider(PredictionStartEndKeepMixin, DataProvider):
+class PredictionProvider(PredictionStartEndKeepMixin, DataProvider[RecordT], Generic[RecordT]):
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/prediction/predictionabc.py
```diff
@@ -249,7 +253,9 @@ async def update_data(
         await self._update_data(force_update=force_update)
 
 
-class PredictionImportProvider(PredictionProvider, DataImportProvider):
+class PredictionImportProvider(
+    PredictionProvider[RecordT], DataImportProvider[RecordT], Generic[RecordT]
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/prediction/priceabc.py
```diff
@@ -1,19 +1,20 @@
...
-from typing import cast
+from typing import Generic, cast
 
 import numpy as np
 import pandas as pd
 from loguru import logger
 from statsmodels.tsa.holtwinters import ExponentialSmoothing
 
 from akkudoktoreos.core.coreabc import PredictionMixin
+from akkudoktoreos.core.dataabc import RecordT
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/prediction/priceabc.py
```diff
@@ -1,19 +1,20 @@
...
 from akkudoktoreos.prediction.predictionabc import PredictionProvider
 from akkudoktoreos.utils.datetimeutil import DateTime, to_datetime, to_duration
 
 
-class PricePredictionProviderBase(PredictionMixin, PredictionProvider):
+class PricePredictionProviderBase(PredictionMixin, PredictionProvider[RecordT], Generic[RecordT]):
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -488,7 +494,7 @@ class DerivedSequence(DataSquence):
     """
 
     # To be overloaded by derived classes.
-    records: list[DataRecord] = Field(
+    records: list[RecordT] = Field(
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -553,7 +559,7 @@ def _validate_key_writable(self, key: str) -> None:
                 f"Key '{key}' is not in writable record keys: {self.record_keys_writable}"
             )
 
-    def _validate_record(self, value: DataRecord) -> None:
+    def _validate_record(self, value: RecordT) -> None:
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -629,7 +635,7 @@ def record_keys_writable(self) -> list[str]:
         return self.record_class().record_keys_writable()
 
     @classmethod
-    def record_class(cls) -> Type:
+    def record_class(cls) -> Type[RecordT]:
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -736,17 +744,18 @@ def __repr__(self) -> str:
...
     # Sequence methods
 
-    def __iter__(self) -> Iterator[DataRecord]:
+    # EOS collections iterate over their elements instead of BaseModel field/value pairs.
+    def __iter__(self) -> Iterator[RecordT]:  # type: ignore[override]
         """Create an iterator for accessing DataRecords sequentially (memory only).
 
         Returns:
-            Iterator[DataRecord]: An iterator for the records.
+            Iterator[RecordT]: An iterator for the records.
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -736,17 +744,18 @@ def __repr__(self) -> str:
...
         """
         return iter(self.records)
 
     async def get_by_datetime(
         self, target_datetime: DateTime, *, time_window: Optional[Duration] = None
-    ) -> Optional[DataRecord]:
+    ) -> Optional[RecordT]:
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -770,7 +779,7 @@ async def get_by_datetime(
 
     async def get_nearest_by_datetime(
         self, target_datetime: DateTime, time_window: Optional[Duration] = None
-    ) -> Optional[DataRecord]:
+    ) -> Optional[RecordT]:
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -800,7 +809,7 @@ async def get_nearest_by_datetime(
 
     # sync rw write access to data sequence, needs locking in case of use in async.
 
-    async def _insert_by_datetime(self, record: DataRecord) -> None:
+    async def _insert_by_datetime(self, record: RecordT) -> None:
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -958,7 +971,7 @@ async def _key_from_series(self, key: str, series: pd.Series) -> None:
 
     # data sequence access usable also for async access
 
-    async def insert_by_datetime(self, record: DataRecord) -> None:
+    async def insert_by_datetime(self, record: RecordT) -> None:
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -1746,7 +1763,7 @@ async def load(self) -> bool:
 # ==================== DataProvider ====================
 
 
-class DataProvider(SingletonMixin, DataSequence):
+class DataProvider(SingletonMixin, DataSequence[RecordT], Generic[RecordT]):
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -2332,7 +2349,7 @@ async def import_from_file(
 # ==================== DataImportProvider ====================
 
 
-class DataImportProvider(DataImportMixin, DataProvider):
+class DataImportProvider(DataImportMixin, DataProvider[RecordT], Generic[RecordT]):
```
b0661: Should be DataRecordT for consistency.

scripts/generate_config_md.py
```diff
@@ -462,6 +467,7 @@ def write_to_file(file_path: Optional[Union[str, Path]], config_md: str):
 
     # Assure timezone name does not leak to documentation
     tz_name = to_datetime().timezone_name
+    assert tz_name is not None
```
b0661: Please also avoid asserts also in scripts (not only regular modules). Raise an RuntimeError,  this will also satisfy mypy.

src/akkudoktoreos/config/config.py
```diff
@@ -790,7 +798,10 @@ def revert_settings(self, backup_id: str) -> None:
                 required by ``self._setup()``.
             OSError: If reading the backup file fails due to I/O issues.
         """
-        backup_file_path = self.general.config_file_path.with_suffix(f".{backup_id}")
+        config_file_path = self.general.config_file_path
+        if config_file_path is None:
+            raise RuntimeError("Configuration file path is not initialized")
```
b0661: At this stage the config file path must be available. If it is not this is a clear implementation issue. Do not mark this a RuntimeError.

src/akkudoktoreos/config/config.py
```diff
@@ -823,7 +834,9 @@ def list_backups(self) -> dict[str, dict[str, Any]]:
         """
         result: dict[str, dict[str, Any]] = {}
 
-        base_path: Path = self.general.config_file_path
+        base_path = self.general.config_file_path
+        if base_path is None:
+            raise RuntimeError("Configuration file path is not initialized")
```
b0661: At this stage the config file path must be available. If it is not this is a clear implementation issue. Do not mark this a RuntimeError.

src/akkudoktoreos/config/configabc.py
```diff
@@ -409,27 +409,31 @@ def available_duration(self, reference_date: Optional[DateTime] = None) -> Optio
         return self.duration
 
 
-class TimeWindowSequence(SettingsBaseModel):
+WindowT = TypeVar("WindowT", bound=TimeWindow)
```
b0661: For consistency this should be TimeWindowT (not WindowT).

src/akkudoktoreos/config/configabc.py
```diff
@@ -409,27 +409,31 @@ def available_duration(self, reference_date: Optional[DateTime] = None) -> Optio
...
+
+
+class TimeWindowSequence(SettingsBaseModel, Generic[WindowT]):
```
b0661: For consistency this should be TimeWindowT (not WindowT).

src/akkudoktoreos/config/configabc.py
```diff
@@ -409,27 +409,31 @@ def available_duration(self, reference_date: Optional[DateTime] = None) -> Optio
...
     """Model representing a sequence of time windows with collective operations.
 
     Manages multiple TimeWindow objects and provides methods to work with them
     as a cohesive unit for scheduling and availability checking.
     """
 
-    windows: list[TimeWindow] = Field(
+    windows: list[WindowT] = Field(
```
b0661: For consistency this should be TimeWindowT (not WindowT).

src/akkudoktoreos/config/configabc.py
```diff
@@ -409,27 +409,31 @@ def available_duration(self, reference_date: Optional[DateTime] = None) -> Optio
...
         default_factory=list,
         json_schema_extra={"description": "List of TimeWindow objects that make up this sequence."},
     )
 
-    def __iter__(self) -> Iterator[TimeWindow]:
+    # EOS collections iterate over their elements instead of BaseModel field/value pairs.
+    def __iter__(self) -> Iterator[WindowT]:  # type: ignore[override]
```
b0661: For consistency this should be TimeWindowT (not WindowT).

src/akkudoktoreos/config/configabc.py
```diff
@@ -409,27 +409,31 @@ def available_duration(self, reference_date: Optional[DateTime] = None) -> Optio
...
         """Allow iteration over the time windows."""
         return iter(self.windows)
 
     def __len__(self) -> int:
         """Return the number of time windows in the sequence."""
         return len(self.windows)
 
-    def __getitem__(self, index: int) -> TimeWindow:
+    def __getitem__(self, index: int) -> WindowT:
```
b0661: For consistency this should be TimeWindowT (not WindowT).

src/akkudoktoreos/config/configabc.py
```diff
@@ -536,7 +540,7 @@ def available_duration(self, reference_date: Optional[DateTime] = None) -> Optio
             total += d
         return total
 
-    def get_applicable_windows(self, reference_date: Optional[DateTime] = None) -> list[TimeWindow]:
+    def get_applicable_windows(self, reference_date: Optional[DateTime] = None) -> list[WindowT]:
```
b0661: For consistency this should be TimeWindowT (not WindowT).

src/akkudoktoreos/config/configabc.py
```diff
@@ -556,7 +560,7 @@ def get_applicable_windows(self, reference_date: Optional[DateTime] = None) -> l
 
     def find_windows_for_duration(
         self, duration: Duration, reference_date: Optional[DateTime] = None
-    ) -> list[TimeWindow]:
+    ) -> list[WindowT]:
```
b0661: For consistency this should be TimeWindowT (not WindowT).

src/akkudoktoreos/config/configabc.py
```diff
@@ -575,7 +579,7 @@ def find_windows_for_duration(
 
     def get_all_possible_start_times(
         self, duration: Duration, reference_date: Optional[DateTime] = None
-    ) -> list[tuple[DateTime, DateTime, TimeWindow]]:
+    ) -> list[tuple[DateTime, DateTime, WindowT]]:
```
b0661: For consistency this should be TimeWindowT (not WindowT).

src/akkudoktoreos/config/configabc.py
```diff
@@ -739,15 +743,15 @@ def to_series(
             dtype=np.float64,
         )
 
-    def add_window(self, window: TimeWindow) -> None:
+    def add_window(self, window: WindowT) -> None:
```
b0661: For consistency this should be TimeWindowT (not WindowT).

src/akkudoktoreos/config/configabc.py
```diff
@@ -739,15 +743,15 @@ def to_series(
...
         """Add a new time window to the sequence.
 
         Args:
             window: The TimeWindow to add.
         """
         self.windows.append(window)
 
-    def remove_window(self, index: int) -> TimeWindow:
+    def remove_window(self, index: int) -> WindowT:
```
b0661: For consistency this should be TimeWindowT (not WindowT).

src/akkudoktoreos/config/configabc.py
```diff
@@ -781,7 +785,7 @@ def sort_windows_by_start_time(self, reference_date: Optional[DateTime] = None)
         if reference_date is None:
             reference_date = pendulum.today()
 
-        def sort_key(window: TimeWindow) -> tuple[int, DateTime]:
+        def sort_key(window: WindowT) -> tuple[int, DateTime]:
```
b0661: For consistency this should be TimeWindowT (not WindowT).

src/akkudoktoreos/core/dataabc.py
```diff
@@ -441,7 +444,10 @@ def keys_from_descriptions(
 # ==================== DataSequence ====================
 
 
-class DataSequence(DataABC, DatabaseRecordProtocolMixin[DataRecord]):
+RecordT = TypeVar("RecordT", bound=DataRecord)
```
b0661: Should be DataRecordT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -2109,8 +2126,8 @@ async def _import_from_json(
 
         # Try dictionary with special keys start_datetime and interval
         try:
-            import_data = PydanticDateTimeData.model_validate_json(json_str)
-            await self._import_from_dict(import_data.to_dict())
+            import_dict = PydanticDateTimeData.model_validate_json(json_str)
```
b0661: Do not make renames. import_data is just what it is named - import data.

src/akkudoktoreos/core/cache.py
b0661, line 788: This is cosmetic. Same logic. Is mypy complaining here?

src/akkudoktoreos/core/dataabc.py
```diff
@@ -1130,7 +1143,11 @@ async def key_to_lists(
                 end_timestamp is None or record_date_time_timestamp < end_timestamp
             ):
                 filtered_records.append(record)
-        dates = [record.date_time for record in filtered_records]
+        dates: list[DateTime] = []
```
b0661: This is completely unnecessary - filtered records always have a date_time that is not None. Suppress the mypy error if it is given here. We do not want to iterate the filtered records once more.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -1378,7 +1395,7 @@ async def key_to_series(
                 start_epoch = int(resample_start.timestamp())
                 floored_epoch = (start_epoch // interval_sec) * interval_sec
 
-                resample_origin: Union[pd.Timestamp, str] = pd.Timestamp(
+                resample_origin: Union[pd.Timestamp, DateTime, str, None] = pd.Timestamp(
```
b0661: How can a pd.Timestamp become None or a DateTime? Do not widen the type here.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -2363,7 +2383,7 @@ class DataContainer(SingletonMixin, DataABC):
     """
 
     # To be overloaded by derived classes.
-    providers: list[DataProvider] = Field(
+    providers: list[ProviderT] = Field(
```
b0661: Should be DataProviderT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -2381,7 +2401,7 @@ def _container_lock(self) -> asyncio.Lock:
             return lock
 
     @field_validator("providers", mode="after")
-    def check_providers(cls, value: list[DataProvider]) -> list[DataProvider]:
+    def check_providers(cls, value: list[ProviderT]) -> list[ProviderT]:
```
b0661: Should be DataProviderT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -2894,7 +2915,7 @@ async def key_delete_by_datetime(
         if key_error:
             raise KeyError(f"key `{key}` is not in predictions")
 
-    def provider_by_id(self, provider_id: str) -> DataProvider:
+    def provider_by_id(self, provider_id: str) -> ProviderT:
```
b0661: Should be DataProviderT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -2350,7 +2367,10 @@ class DataImportProvider(DataImportMixin, DataProvider):
 # ==================== DataContainer ====================
 
 
-class DataContainer(SingletonMixin, DataABC):
+ProviderT = TypeVar("ProviderT", bound=DataProvider)
```
b0661: Should be DataProviderT for consistency.

src/akkudoktoreos/core/dataabc.py
```diff
@@ -2350,7 +2367,10 @@ class DataImportProvider(DataImportMixin, DataProvider):
...
+
+
+class DataContainer(SingletonMixin, DataABC, Generic[ProviderT]):
```
b0661: Should be DataProviderT for consistency.

src/akkudoktoreos/prediction/loadabc.py
```diff
@@ -20,7 +20,10 @@ class LoadDataRecord(PredictionRecord):
     )
 
 
-class LoadProvider(PredictionProvider):
+LoadRecordT = TypeVar("LoadRecordT", bound=LoadDataRecord)
```
b0661: This should be LoadDataRecordT for consistency.

src/akkudoktoreos/prediction/loadabc.py
```diff
@@ -20,7 +20,10 @@ class LoadDataRecord(PredictionRecord):
...
+
+
+class LoadProvider(PredictionProvider[LoadRecordT], Generic[LoadRecordT]):
```
b0661: This should be LoadDataRecordT for consistency.

src/akkudoktoreos/prediction/loadabc.py
```diff
@@ -41,7 +44,7 @@ class LoadProvider(PredictionProvider):
     """
 
     # overload
-    records: List[LoadDataRecord] = Field(
+    records: List[LoadRecordT] = Field(
```
b0661: This should be LoadDataRecordT for consistency.

src/akkudoktoreos/prediction/weatherclearoutside.py
```diff
@@ -336,9 +337,9 @@ async def _update_data(self, force_update: Optional[bool] = None) -> None:
                     if key is None:
                         continue
                     if detail_name in clearout_data:
-                        value = clearout_data[detail_name][row_index]
+                        record_value = clearout_data[detail_name][row_index]
```
b0661, line 340: What is the problem with value? Why using record_value now?

src/akkudoktoreos/server/dash/configuration.py
```diff
@@ -250,7 +251,7 @@ def resolve_nested_types(field_type: Any, parent_types: list[str]) -> list[tuple
 
 
 def create_config_details(
-    model: type[PydanticBaseModel], values: dict, values_prefix: list[str] = []
+    model: type[BaseModel], values: dict, values_prefix: list[str] = []
```
b0661: PydanticBaseModel and BaseModel are diofferent types. PydanticBaseModel is derived from BaseModel and provides some extras. Do not widen the type.

b0661 (changes requested):
Thanks for the PR.

A first batch of review comments. A lot is formal. At some places the changes does not make sense in my view.

src/akkudoktoreos/adapter/homeassistant.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. The type check now raises `TypeError` outside the broad exception handlers. Simply replacing `return []` with a raise inside the original `try` would still have swallowed the implementation error. All three computed properties have regression coverage for a misregistered provider, a valid provider, and the existing unavailable-provider/service fallbacks.

src/akkudoktoreos/adapter/homeassistant.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. This property now raises `TypeError` for a misregistered provider, outside the existing unavailable-service fallback. Covered by the same parametrized regression tests as the other two entity-list properties.

src/akkudoktoreos/adapter/homeassistant.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. This property now raises `TypeError` for a misregistered provider, outside the existing unavailable-service fallback. Covered by the same parametrized regression tests as the other two entity-list properties.

src/akkudoktoreos/core/cache.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Replaced this with an explicit `raise AssertionError` and documented that the preceding initialization guarantees the threshold whenever `clear_all` is false. This describes an internal invariant and remains active with Python optimization; no `assert` statement is used.

src/akkudoktoreos/core/databaseabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. `UTC` is explicitly re-exported by `datetimeutil`, and `databaseabc` imports it from there. The five timestamp factory calls retain the timezone object they require; the direct Pendulum import has been removed from this production module.

src/akkudoktoreos/core/databaseabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. `DatabaseTimestamp.from_datetime` again accepts only `DateTime`. The record protocol constructor also now takes `DateTime` instead of `Any`. Optional record values are narrowed at the database boundary or after an established invariant. The defensive runtime rejection of invalid input remains covered without widening the public annotation.

src/akkudoktoreos/core/databaseabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. The constructor parameter is now `DateTime`, and `DatabaseTimestamp.from_datetime` is strict again. The protocol field itself remains optional for compatibility: `DataRecord.date_time` was already `Optional[DateTime] = None` before this PR, and existing record tests explicitly construct incomplete records and delete their timestamps. `DateTime` is already a concrete Pendulum type for mypy in `datetimeutil`. Making only the protocol field non-optional would misrepresent the implementing models.

Open design question: should the database get a separate interface for records whose timestamp has been validated, or is preserving the optional in-memory record contract with narrowing at the database boundary acceptable here? This follow-up uses the latter and avoids changing public record construction/serialization.

src/akkudoktoreos/core/databaseabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. The helper is now used only at database entry points (insert, deserialization/load, and dirty tracking), where an incoming record can actually be incomplete. It is no longer called repeatedly during nearest-record searches or compaction. Those sites use documented casts after the relevant invariant/filter. The remaining optional-field design question is explained in [the protocol discussion](https://github.com/Akkudoktor-EOS/EOS/pull/1284#discussion_r3951252705).

src/akkudoktoreos/core/databaseabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Removed the runtime helper call from this search. A documented cast describes the timestamp invariant established when records enter the index. The optional in-memory record contract is retained; see [the protocol discussion](https://github.com/Akkudoktor-EOS/EOS/pull/1284#discussion_r3951252705).

src/akkudoktoreos/core/databaseabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Removed the runtime helper call from the distance check. The indexed record timestamp is narrowed with a cast under the documented insertion/loading invariant. See [the protocol discussion](https://github.com/Akkudoktor-EOS/EOS/pull/1284#discussion_r3951252705) for the remaining model-contract question.

src/akkudoktoreos/core/databaseabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. This call now receives `UTC` through `akkudoktoreos.utils.datetimeutil`; `databaseabc` no longer imports Pendulum directly. Timestamp flooring and UTC semantics are unchanged.

src/akkudoktoreos/core/databaseabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. This call now receives `UTC` through `akkudoktoreos.utils.datetimeutil`; `databaseabc` no longer imports Pendulum directly. Timestamp flooring and UTC semantics are unchanged.

src/akkudoktoreos/core/databaseabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Removed the repeated runtime helper calls from sorting and bucket construction. The preceding window filter already excludes missing timestamps; documented casts carry that knowledge into these expressions. See [the protocol discussion](https://github.com/Akkudoktor-EOS/EOS/pull/1284#discussion_r3951252705) for why the general record field remains optional.

src/akkudoktoreos/core/databaseabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Removed the runtime helper call here. The preceding window filter excludes missing timestamps, so the expression now uses a cast under that documented invariant. See [the protocol discussion](https://github.com/Akkudoktor-EOS/EOS/pull/1284#discussion_r3951252705) for the remaining model-contract question.

src/akkudoktoreos/core/databaseabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. This call now receives `UTC` through `akkudoktoreos.utils.datetimeutil`; `databaseabc` no longer imports Pendulum directly. Timestamp flooring and UTC semantics are unchanged.

src/akkudoktoreos/core/databaseabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. This call now receives `UTC` through `akkudoktoreos.utils.datetimeutil`; `databaseabc` no longer imports Pendulum directly. Timestamp flooring and UTC semantics are unchanged.

src/akkudoktoreos/core/ems.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Removed the redundant failure branch entirely: when initialization is needed, the property returns `EnergyManagement.set_start_datetime()` directly, whose return type is already `DateTime`. A deterministic regression test verifies hour rounding and that the clock is read only on first access.

src/akkudoktoreos/prediction/predictionabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/prediction/predictionabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/prediction/predictionabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/prediction/priceabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/prediction/priceabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

scripts/generate_config_md.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Replaced the `assert` with an explicit `RuntimeError` for a missing timezone name. A regression test verifies that generation fails before writing the output file.

src/akkudoktoreos/config/config.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Changed the guard to an explicit `raise AssertionError` with a comment documenting the configuration-setup invariant. This avoids classifying it as an ordinary runtime condition, without using an `assert` statement that disappears under `python -O`. Both backup operations have regression coverage.

src/akkudoktoreos/config/config.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Changed the guard to an explicit `raise AssertionError` and documented the configuration-setup invariant, matching `revert_settings`. Both backup operations have regression coverage.

src/akkudoktoreos/config/configabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `TimeWindowT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/config/configabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `TimeWindowT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/config/configabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `TimeWindowT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/config/configabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `TimeWindowT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/config/configabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `TimeWindowT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/config/configabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `TimeWindowT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/config/configabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `TimeWindowT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/config/configabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `TimeWindowT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/config/configabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `TimeWindowT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/config/configabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `TimeWindowT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/config/configabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `TimeWindowT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Restored `import_data` for the dictionary-model branch. The original unannotated reuse causes a mypy assignment error because the first branch establishes `PydanticDateTimeDataFrame` and the next assigns `PydanticDateTimeData`. An explicit union now describes the supported import representations; the raw JSON boundary is narrowed to the dictionary expected by the importer. Import behavior is preserved.

src/akkudoktoreos/core/cache.py
dr-dimitri: Rechecked with a minimal example using the configured mypy plugins: yes, the original reassignment produces `Incompatible types in assignment (expression has type "float", variable has type "Duration | None")`. The conditional expression lets mypy infer the actual result without reusing a duration-typed variable for seconds. The runtime logic, including the zero-duration case, is unchanged. I have therefore retained this part of the PR.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Removed the repeated `None` check and restored the list comprehension, with one cast of its result and a comment explaining the preceding filter. No additional traversal is introduced. For precision, the previous PR loop replaced a comprehension, so the issue was redundant per-record validation rather than an extra complete traversal.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Restored `Union[pd.Timestamp, str]`. The other branches previously assigned `resample_start`, which can be `DateTime` or `None`; that was the reason for the wider annotation. They now convert a supplied origin to `pd.Timestamp` and use `"start_day"` when no origin exists, preserving the bucket origin while keeping the type narrow. The sequence/resampling regression suite passes.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataProviderT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataProviderT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataProviderT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataProviderT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/core/dataabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `DataProviderT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/prediction/loadabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `LoadDataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/prediction/loadabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `LoadDataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/prediction/loadabc.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Renamed this type variable to `LoadDataRecordT` consistently across its definition, imports, annotations, and affected docstrings. The generic bounds are unchanged.

src/akkudoktoreos/prediction/weatherclearoutside.py
dr-dimitri: Rechecked with mypy: `value` is already declared `float | str` earlier in the same function when parsing HTML text. Reusing that name here makes multiplication by the floating correction factor fail with `Unsupported operand types for * ("str" and "float")`. `record_value` keeps the later heterogeneous record-building step separate from that parsed-text variable; it also handles the timestamp column. This does not claim that the existing `dict[str, Any]` boundary is fully statically checked. I have retained the rename because it addresses a reproduced type error without changing runtime conversion behavior.

src/akkudoktoreos/server/dash/configuration.py
dr-dimitri: Prepared in the local follow-up patch; not committed or pushed yet. Removed the general `BaseModel` parameter. The annotation now explicitly accepts `type[PydanticBaseModel] | type[ConfigEOS]`. The second case is necessary because the existing caller passes `ConfigEOS`, which inherits through `SettingsEOS` from `BaseSettings`, not from `PydanticBaseModel`; restoring only the old annotation reproduces a mypy argument error. A regression test verifies matching details for the top-level settings and a nested EOS model.

dr-dimitri:
I rechecked all 61 review comments and published the follow-up patch. Each discussion now has an explanation of the change or the reason for retaining the existing code.

**Publication status:** the follow-up changes are now committed and pushed to `fix/typing-debt` in [commit a043079](https://github.com/Akkudoktor-EOS/EOS/pull/1284/commits/a043079e19c97842a2eba1906593af93a0b887a7), verified as the current PR head. The earlier inline replies described the patch before publication; those changes are now available in the PR. Review threads remain open for reviewer confirmation, including the design question below.

The patch applies the consistent generic names, routes UTC through `datetimeutil`, makes misregistered Home Assistant providers raise visibly, removes the redundant EMS failure branch, clarifies internal invariant failures, removes the script assertion, restores `import_data`, removes redundant timestamp checks, keeps resampling origins narrow, and limits the dashboard model argument to the EOS types it actually receives.

The remaining design question is in [the record protocol discussion](https://github.com/Akkudoktor-EOS/EOS/pull/1284#discussion_r3951252705): the existing in-memory record model allows an absent timestamp, while the database requires one. The patch restores a strict `from_datetime(DateTime)` parameter, tightens the protocol constructor, validates at database entry points, and narrows values after established invariants. It preserves the existing record model contract pending review of whether a separate interface for validated records is preferred.

The cache expression and weather variable rename are retained with reproduced mypy diagnostics explained in their discussions.

Validation of the published patch:

- Relevant regression suite with `--finalize --check-config-side-effect`: **748 passed, 5 skipped**. The 3 generated-artifact freshness tests were deselected; the comparison below instead verifies that this patch introduces no additional generated-artifact drift. The initial sandbox-only DNS failure passes when DNS access is available.
- Full `src`/`tests` mypy without incremental cache: **passes, 220 files**.
- `make format`: **all hooks pass**, including the isolated mypy hook.
- `make gen-docs` and a separate regeneration of unchanged PR HEAD: all **20 configuration documents match**, and OpenAPI matches apart from the generated version. Existing generator/version drift was kept out of the patch.

This is a targeted regression run, not a new full `make test-ci` run or Linux CI result.

PR #1305, opened 2026-09-13
fix: runtime config updates discarded by config file and env

### LINKED ISSUE #1303: PUT /v1/config returns 200 OK but does not persist changes to nested list/model fields (pvforecast.planes, optimization.genetic)
### Describe the bug
A bulk `PUT /v1/config` request that includes a full `pvforecast` section (with `planes`) or `optimization.genetic` section returns HTTP 200 with the (apparently) updated config in the response, but a subsequent `GET /v1/config` shows the old/default values unchanged.

### Environment
- EOS version: 0.3.0.dev2608221553448643
- Installation: Home Assistant Add-on

### Steps to Reproduce
1. `PUT /v1/config` with body:
```json
{
  "optimization": {"algorithm": "GENETIC", "genetic": {"individuals": 60, "generations": 60}},
  "pvforecast": {"provider": "PVForecastAkkudoktor", "planes": [{"surface_tilt": 25.0, "surface_azimuth": 117.0, "peakpower": 9.775}], "max_planes": 1}
}
```
2. Response: `200 OK`.
3. `GET /v1/config` immediately after: `optimization.genetic.individuals` still `200` (previous value), `pvforecast.planes` still contains the original 4 example/demo planes (`Sungrow__SH25T`, `AXITEC_AC_410MH_144S`, azimuths 170/90/140/185) instead of the submitted single plane.

### Workaround found
The granular per-field endpoint works reliably:
```
PUT /v1/config/optimization/genetic/individuals   body: 60
PUT /v1/config/optimization/genetic/generations   body: 60
PUT /v1/config/pvforecast/planes                  body: [<single plane object>]
```
Each of these, verified via a follow-up `GET`, correctly updates and persists. Only the bulk `PUT /v1/config` path with these nested structures silently fails to apply.

### Expected behavior
Bulk `PUT /v1/config` should apply and persist nested list/model field updates the same way the granular `PUT /v1/config/{path}` endpoint does.

### Additional context
This caused real-world impact: unnoticed, EOS kept running its genetic optimization against 4 demo PV planes and `individuals=200/generations=200` instead of the intended single-plane, reduced-workload config, contributing to memory pressure and repeated OOM-kills (exit code 137) on a resource-constrained host.

r0b2g1t:
**Problem**

`PUT /v1/config` returns 200 but does not apply nested updates (`pvforecast.planes`, `optimization.genetic`). `PUT /v1/config/{path}` applies the same values correctly (#1303).

**Cause**

`merge_settings_from_dict()` passes the update to `_setup(**merged)`, i.e. as `pydantic-settings init settings`. Since #1234 those rank below the EOS configuration file and the environment, so any key present there discarded the update. The granular endpoint is unaffected because set_nested_value() mutates the model in place without re-evaluating the sources.

`revert_settings()` and `update()` used the same path and were broken the same way.

Second defect found while tracing: `ConfigEOS.__init__` passed `self` as first positional argument to `_setup()`, which forwards it to `BaseSettings.__init__`, where it lands in _case_sensitive. Environment lookup became case sensitive and returned nothing on the first build — the very behaviour #1234 tried to fix by reordering the sources.

**Changes**

- New runtime settings source, ranked below the CLI arguments and above environment, `dotenv` and config file. Restores the priority documented in `configuration.md`.
- `merge_settings_from_dict()` stores only the explicitly provided values there, so environment changes stay effective for keys not set at runtime.
- `set_nested_value()` records into the same store, so granular updates survive a later bulk update. List index paths store the whole list, cleared values store `None`.
- `reset_settings()` clears the store; `revert_settings()` and `update()` go through it.
- `_setup()` no longer receives `self`, so environment variables apply on the first build.
- Priority list in the docs plus changelog entry.

**Tests**

Regression tests in `tests/test_config.py`: bulk update over config file and overstill beating the config file for untouched keys (guards #1234), reset, granular updates surviving a bulk update (list index, cleared value), revert_settings(), environment on first build.

Verified against the endpoint with the reported config file:

GET before : individuals 200, planes 4
PUT        : 200, individuals 300, planes 1
GET after  : individuals 300, planes 1
POST reset : individuals 200, planes 4

Closes #1303


b0661:
Thank you, pydantic is a really beast. Please refresh code formatting to make pre-commit pass the code check.

NormannK:
i was about to merge it but there is a small problem imo.

`revert_settings()` currently serializes the migrated backup with `exclude_defaults=True`. This drops values that are explicitly present in the backup but happen to equal their model defaults.

For example, if the backup contains `server.port: 8503` while the active config file specifies `9000`, the restored runtime settings omit `8503`, allowing the config-file value to remain active. The endpoint therefore reports a successful revert without fully restoring the backup.

Suggested change:

`exclude_none=True, exclude_unset=True, exclude_computed_fields=True`

instead of:

`exclude_none=True, exclude_defaults=True, exclude_computed_fields=True`

r0b2g1t:
Good catch, fixed in the last commit — different mechanism though, `exclude_unset=True` does not work here.

`migrate_config_data()` fills the nested settings of a fresh `SettingsEOSDefaults()` in place, so the root model has no field set at all:

```
root fields_set   : []
server fields_set : ['port']
unset dump        : {}
```

`exclude_unset=True` drops every section at the root, so a revert would restore nothing.

Instead, a new `dump_set_fields()` helper walks the nested models and keeps every explicitly set value, default-valued or not:

```
ConfigEOS._runtime_settings = dump_set_fields(
    backup_settings, exclude_none=True, exclude_computed_fields=True
)
```

Your example is now in `test_revert_settings_restores_backup`: backup `server.port: 8503` over config file `9000` restores `8503`. Lists of nested models restore too (checked with `pvforecast.planes`). Suite green, mypy clean.

Note: `to_config_json()` still writes with `exclude_defaults=True`, so an EOS-written backup never contains a default-valued key in the first place. Existing behaviour on main, own PR.

PR #1332, opened 2026-09-17
fix(server): redirect to EOSdash at the client reachable host

### LINKED ISSUE #1320: EOSdash of Home Assistant add-on is not accessible through an external reverse proxy
### Describe the issue

I am running the Akkudoktor-EOS Home Assistant add-on behind an external Traefik reverse proxy with Authentik forward authentication.

The EOS REST API on port 8503 is reachable without problems. However, opening the root endpoint of the EOS REST API returns an absolute redirect to localhost:

```http
HTTP/1.1 303 See Other
Location: http://127.0.0.1:8504/
```

The browser follows this redirect and tries to connect to port 8504 on the client computer itself.

EOSdash is initially started with:

```text
Starting EOSdash server on 127.0.0.1:8504
Uvicorn running on http://127.0.0.1:8504
```

Changing the EOS configuration setting:

```json
{
  "server": {
    "eosdash_host": "0.0.0.0"
  }
}
```

successfully changes the EOSdash bind address:

```text
Starting EOSdash server on 0.0.0.0:8504
Uvicorn running on http://0.0.0.0:8504
```

However, port 8504 is not published by the Home Assistant add-on. Only port 8503 is exposed on the Home Assistant host.

```text
$ docker port app_ec9bffb2_eos

8503/tcp -> 0.0.0.0:8503
8503/tcp -> [::]:8503
```

Therefore, the external reverse proxy cannot connect to:

```text
http://HOME_ASSISTANT_HOST:8504
```

The connection is refused:

```text
curl http://HOME_ASSISTANT_HOST:8504/

curl: (7) Failed to connect to HOME_ASSISTANT_HOST port 8504:
Could not connect to server
```

Home Assistant Ingress works, but direct access through an authenticated external reverse proxy is not possible without using the internal add-on container IP.

### Environment

- Home Assistant OS
- Akkudoktor-EOS Home Assistant add-on
- Add-on image version: `0.3.0.dev2608221553448643`
- Architecture: `amd64`
- Base image: `debian:trixie-slim`
- Base image version: `2026.08.0`
- External reverse proxy: Traefik
- Authentication middleware: Authentik
- EOS REST API port: `8503`
- EOSdash port: `8504`
- Home Assistant Ingress: enabled

Running image:

```text
ec9bffb2/amd64-addon-eos:0.3.0.dev2608221553448643
```

Relevant image labels:

```json
{
  "io.hass.arch": "amd64",
  "io.hass.base.image": "debian:trixie-slim",
  "io.hass.base.version": "2026.08.0",
  "io.hass.name": "Akkudoktor-EOS",
  "io.hass.type": "app",
  "io.hass.version": "0.3.0.dev2608221553448643"
}
```

### Steps to reproduce

1. Install and start the Akkudoktor-EOS Home Assistant add-on.
2. Configure an external reverse proxy to access the EOS REST API at:

```text
http://HOME_ASSISTANT_HOST:8503
```

3. Open the externally configured EOS hostname.
4. EOS responds with:

```http
HTTP/1.1 303 See Other
Location: http://127.0.0.1:8504/
```

5. The browser follows the redirect and tries to access port 8504 on its own localhost.
6. The connection fails.
7. Set `server.eosdash_host` to `0.0.0.0`.
8. Restart the add-on.
9. EOSdash now listens on `0.0.0.0:8504`, but port 8504 is still not published on the Home Assistant host.
10. Access to `http://HOME_ASSISTANT_HOST:8504` continues to fail with `Connection refused`.

### Current workaround

The current workaround is to obtain the internal add-on container IP:

```bash
docker inspect app_ec9bffb2_eos \
  --format '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{println}}{{end}}'
```

Example result:

```text
172.30.33.4
```

Traefik can then connect directly to the internal container address:

```yaml
http:
  routers:
    eos:
      rule: "Host(`eos.example.com`)"
      service: eos
      entryPoints:
        - websecure
      tls:
        certResolver: cloudflare
      middlewares:
        - authentik@file

  services:
    eos:
      loadBalancer:
        passHostHeader: true
        servers:
          - url: "http://172.30.33.4:8504"
```

EOSdash works correctly through Traefik and Authentik with this workaround.

However, the workaround is fragile because the internal add-on container IP may change after an add-on update or container recreation.

### Expected behavior

EOSdash should optionally be accessible through an external reverse proxy without requiring the internal and potentially changing add-on container IP.

Possible solutions:

1. Optionally publish port 8504 in the Home Assistant add-on configuration:

```yaml
ports:
  8503/tcp: 8503
  8504/tcp: 8504

ports_description:
  8503/tcp: "EOS REST server"
  8504/tcp: "EOSdash dashboard server"
```

2. Allow `server.eosdash_host` to be configured through the Home Assistant add-on options.

3. Avoid redirecting external clients to:

```text
http://127.0.0.1:8504/
```

4. Use a relative redirect or respect forwarded headers such as:

```text
X-Forwarded-Host
X-Forwarded-Proto
X-Forwarded-Port
```

5. Alternatively, provide a configurable external EOSdash URL, for example:

```text
server.eosdash_external_url
```

6. Keep Home Assistant Ingress enabled while allowing users to optionally expose port 8504.

### Actual behavior

- EOS REST API is externally reachable on port 8503.
- The root endpoint redirects the browser to `127.0.0.1:8504`.
- EOSdash listens on `127.0.0.1:8504` by default.
- Changing `eosdash_host` to `0.0.0.0` makes EOSdash listen on the container network.
- Port 8504 remains unavailable through the Home Assistant host.
- An external reverse proxy must use the internal add-on container IP.

### Security considerations

Publishing port 8504 should remain optional.

Users who only use Home Assistant Ingress should be able to retain the current secure default. Users operating an authenticated reverse proxy should be able to explicitly enable port 8504 in the add-on network configuration.

In this setup, external access is protected by:

- HTTPS through Traefik
- Authentik forward authentication
- No direct public exposure of the Home Assistant host ports

### Additional context

The EOS REST API and EOSdash are both functioning correctly.

The issue is specifically caused by the combination of:

- EOSdash binding to `127.0.0.1` by default
- the absolute redirect to `http://127.0.0.1:8504/`
- port 8504 not being optionally exposed by the Home Assistant add-on

Once Traefik connects directly to the internal container address on port 8504, EOSdash works correctly.

r0b2g1t:
Remote browsers were redirected to EOSdash's bind address (often their own localhost), and the original fix split IPv6 addresses incorrectly and assumed that a reverse proxy exposed the internal dashboard port.

This change uses the request hostname for direct dashboard access, preserving IPv6, and adds optional `server.eosdash_public_url` for the externally reachable dashboard base URL. For example, `https://energy.example.com:9443/dashboard` preserves the public scheme, port and path prefix for both redirects and the dashboard link on 404 pages. It does not change bind addresses or configure proxy routing.

Raw `X-Forwarded-Host` / `X-Forwarded-Proto` headers do not override the destination; ordinary scheme handling follows the ASGI server's trusted-proxy configuration. Public URLs reject credentials, queries, fragments and invalid ports.

The configuration schema and API documentation are regenerated. Regression coverage includes direct IPv4/hostname/IPv6 access, TLS termination, public ports, path prefixes, 404 links and invalid settings.

Validation: 23 focused redirect/configuration cases passed; mypy passed for all 268 source files. Formatting and documentation checks passed. The broader local server/configuration run finished with 74 passed, 1 skipped, 1 failed and 2 setup errors: the three Windows process-start cases failed on expected-versus-reported process IDs/startup readiness. These are not claimed as passing. Fresh Linux CI checks are running on this PR.

Closes #1320


tests/test_server.py
```diff
@@ -369,3 +371,62 @@
...
+        assert response.headers["location"] == "http://eos.example.com:8504/eosdash/health"
+
+    def test_unknown_path_error_page_links_to_request_host(self, client):
+        """The 404 page links to EOSdash on the host the client used."""
+        response = client.get("/no-such-page", headers={"Host": "eos.example.com"})
+        assert response.status_code == HTTPStatus.NOT_FOUND
+        # Compare whole link targets, a substring check would also accept a foreign host.
+        # The error message is HTML escaped by the error page, so unescape it first.
+        hrefs = re.findall(r'href="([^"]*)"', html.unescape(response.text))
+        assert "http://eos.example.com:8504/" in hrefs
```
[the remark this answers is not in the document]
r0b2g1t: false positive — a test asserting an expected URL is not a sanitizer

src/akkudoktoreos/server/eos.py
```diff
@@ -2376,20 +2377,36 @@ def _sanitize_redirect_path(path: str) -> Optional[str]:
...
+    if settings.eosdash_public_url:
+        return settings.eosdash_public_url.rstrip("/")
+    port = settings.eosdash_port or 8504
+    scheme = request.url.scheme
+    if scheme not in ("http", "https"):
+        scheme = "http"
+    # Request.url honours the Host header and parses bracketed IPv6 correctly.
+    # Proxy scheme handling belongs to the ASGI server's trusted proxy middleware;
+    # do not trust arbitrary raw X-Forwarded-* headers here.
+    host = urlsplit(str(request.url)).hostname or ""
```
NormannK, line 2423: This makes the endpoint an open redirect because the request `Host` header is untrusted and the app does not install `TrustedHostMiddleware`. For example, `Host: attacker.example` produces a redirect to `http://attacker.example:8504/`.

Please establish a trust boundary before reflecting this value. Options include requiring `eosdash_public_url` for external deployments, adding/configuring `TrustedHostMiddleware` with an explicit allowlist, or falling back to a configured public host unless the request host has been validated. Add a regression test using an untrusted Host value.

tests/test_server.py
```diff
@@ -369,3 +371,108 @@ def test_server_setup_for_class(self, server_setup_for_class):
...
+        assert response.status_code == HTTPStatus.SEE_OTHER
+        assert response.headers["location"] == "http://eos.example.com:8504/eosdash/health"
+
+    def test_unknown_path_error_page_links_to_request_host(self, client):
+        """The 404 page links to EOSdash on the host the client used."""
+        response = client.get("/no-such-page", headers={"Host": "eos.example.com"})
+        assert response.status_code == HTTPStatus.NOT_FOUND
+        # Compare whole link targets, a substring check would also accept a foreign host.
+        # The error message is HTML escaped by the error page, so unescape it first.
+        hrefs = re.findall(r'href="([^"]*)"', html.unescape(response.text))
```
NormannK: Calling `html.unescape(response.text)` makes this test pass even though the browser receives `&lt;a ...&gt;` and displays it as text: `create_error_page()` applies `html.escape(error_message)` to the complete message.

Please test the actual HTML without unescaping it. A safe fix would be to give `create_error_page` a separate link URL/label slot, escape the URL with `html.escape(url, quote=True)`, and keep ordinary error text escaped. Avoid marking the entire interpolated message as safe HTML.

NormannK (changes requested):
Thanks for the thorough tests and documentation. The CodeQL alert on `tests/test_server.py:431` is a false positive: it is analyzing a test assertion against an expected URL, not production sanitization or validation.

I found three issues that should be addressed before merge:

1. **The Home Assistant scenario in #1320 is still not reachable.** The add-on continues to expose only port 8503; port 8504 remains commented out in `config.yaml`, and this PR intentionally adds no proxy route. `eosdash_public_url` changes the browser's destination but does not give Traefik a stable backend route to EOSdash, so the reporter still needs the add-on container IP. Please either expose 8504 as an optional add-on port, proxy EOSdash through the exposed EOS server, or treat this as a partial redirect fix rather than closing #1320.

2. **The request `Host` is reflected into an absolute redirect without host validation.** See the inline comment for mitigation options.

3. **The 404 “link” is escaped and therefore rendered as text.** The test currently unescapes the response and hides that behavior. See the inline comment for a safe implementation approach.

The first issue blocks the stated Home Assistant use case; the other two affect security and user-visible behavior.

NormannK:
The latest patch addresses my two inline points: direct redirects now accept only known request hosts (with an untrusted-Host regression test), and the 404 page emits an escaped, real link that is checked in the raw response HTML. I resolved those threads.

The remaining pytest failures are `test_addon_configuration_complete` and `test_ingress_configuration_consistent` in `tests/test_homeassistant.py`. Both assert that the ingress port cannot appear *at all* in `ports`. The proposed configuration has `ingress_port: 8504` and `8504/tcp: null`; per Home Assistant's app configuration docs, a null host port disables the mapping by default. Thus these two assertions reject the intended optional mapping without checking its default exposure.

Suggested test fix: replace the “ingress port key must be absent” assertion in both tests with a check that `ports[f"{ingress_port}/tcp"] is None` (and explicitly check that the key exists, if optional publication is part of the intended behavior). Keep the ingress-port validity checks. This verifies that ingress remains configured and EOSdash has no host mapping by default; a user can opt in by assigning a host port in the add-on Network settings. Avoid merely deleting the assertions.

PR #1333, opened 2026-09-21
fix(optimization): let the AC setpoint cap the total charge where the inverter does

drbacke:
Some hybrid inverters (e.g. Deye in time-of-use grid charging) limit the whole
battery charge current to the grid charge setpoint. Observed on a Deye 12K:
max_grid_charge_current = 75 A -> ~80 A charge at 8 kW PV, ~3.3 kW exported,
including the surplus of grid-side micro inverters. EOS modelled AC slots as
"PV first, grid adds ac_charge x remaining power", so under PV surplus AC looked
at least as good as DC and the optimizer picked a low AC factor that blocked PV.

Changes
- New inverter config option `devices.inverters[].ac_charge_limits_total_charge`
  (default False -> existing model unchanged).
- When True, an AC slot caps the battery's raw charge at
  ac_charge x max_charge_power_w from all sources; PV above the cap is exported,
  the grid only fills what PV leaves.
- POST /optimize: the request field is optional. If unset, the value comes from
  the configured inverter with the same device_id (or the only configured one);
  an explicit request value still wins.
- AC slot logic moved into Inverter (ac_charge_factor, begin_ac_charge_slot,
  charge_battery_from_grid), shared by the simulation and the tail value curve.
- Battery gets a per-slot charge cap, lifted by reset().
- Regenerated configdevices.md and openapi.json.

Verified live: after enabling the option, the optimizer no longer plans low AC
factors under PV surplus; AC slots now use factors 0.8-1.0 only where it
deliberately buys grid energy.

Tests: tests/test_ac_charge_total_limit.py (inverter level, GeneticSimulation,
config resolution).


src/akkudoktoreos/prediction/pvforecastakkudoktorlocal.py
```diff
@@ -29,26 +29,21 @@
...
 from akkudoktoreos.config.configabc import SettingsBaseModel
 from akkudoktoreos.core.cache import cache_in_file
-from akkudoktoreos.prediction.pvforecastabc import (
-    PVForecastDataRecordT,
-    PVForecastProvider,
-)
+from akkudoktoreos.prediction.pvforecastabc import PVForecastProvider
 from akkudoktoreos.utils.datetimeutil import compare_datetimes, to_datetime, to_duration
 
 OPENMETEO_URL = "https://api.open-meteo.com/v1/forecast"
```
b0661, line 49: Why do you call OpenMeteo directly? Better use the weather data that is stored from the weather prediction (see e.g. PVForecastPVLIB).

tests/test_feedintariffakkudoktor.py
```diff
@@ -15,14 +16,15 @@
 def provider(config_eos):
     config_eos.merge_settings_from_dict(
         {
+            "elecprice": {"charges_kwh": 0.30},
```
b0661, line 19: We have ElecFee providers. Use the fees provided by them  instead of another configuration value.

b0661 (changes requested):
The pull request contains a lot of changes already made to main. Also some concepts (like ElecFee and usage of prediction data by other predictions) are ignored or contradicted. SHould be cleaned up before merge.

NormannK:
The AC charge cap fix described here was ported to the current GENETIC configuration and merged into `main` in #1334. This branch diverged before the GENETIC integration and contains unrelated prediction, fee, and generated-file changes; merging `main` into it produces conflicts across those areas. Closing this PR as superseded by #1334 rather than carrying those older changes into `main`.

